PR #144, opened 2025-07-03
Fix geometry extension

JacsonGilVargas:


andrefpf:
In main window we use this list of valid formats in many places (as shown in the figure bellow).  I think the need to change the supported extensions in multiple places is very error prone.
image

What do you think of creating global variables containing all the supported extensions? We could add the two following variables in `vibra.__init__.py`:

```python
SUPPORTED_GEOMETRY_EXTENSIONS = [
    "iges",
    "IGES",
    "igs",
    "IGS",
    "step",
    "STEP",
    "stp",
    "STP",
]

SUPPORTED_MESH_EXTENSIONS = [
    "bdf",
    "BDF",
    "nas",
    "NAS",
]
```

The Qt inputs might need some formating, but it should be as simple as: 
```python
extensions = " ".join(f"*.{extension}" for extension in SUPPORTED_GEOMETRY_EXTENSIONS)
ext_filter = f"Geometry Files ({extensions})"
```

andrefpf (approved):

PR #147, opened 2025-07-04
Duplicate fluid and material

JacsonGilVargas:
This pull request implements the duplicate fluid/material functionality. Please @andrefpf, execute some tests and check if everything works well.


andrefpf:
There is something wrong with the material exclusions. The selected material is not removed neither when I press "delete" or when I use the "-" button. Instead, the software removes either the first or the neighbor material.

The fluids removal still works exactly as I was expecting.

andrefpf:
I am not sure how hard is it to move the scroll bar to the end, but as the materials/fluids are added to the end of the list, this will be a nice addition.

andrefpf:
> There is something wrong with the material exclusions. The selected material is not removed neither when I press "delete" or when I use the "-" button. Instead, the software removes either the first or the neighbor material.
> 
> The fluids removal still works exactly as I was expecting.

Also, sometimes it raises the following error:
image

andrefpf (approved):
Seems to be working well now

PR #148, opened 2025-07-09
Improve efficiency of frequency dependent acoustic assembler

andrefpf:
Avoid the recreation of csr_matrix when only the data values need to be updated. The data update require a preprocessing step, but for typical configurations it should be advantageous.


JacsonGilVargas:
After a few tests, it was observed that the total time spent on the solution has decreased by up to 8% in "heavy models". From a mathematical perspective, the proposed algorithms worked as intended, but the results we achieved fell short of our initial expectations. Based on these facts, we have decided not to change the structure for creating and updating sparse matrices.

PR #153, opened 2025-07-18
Color class in renderers

GuilhermeSPierri:
This pull request:

- Replaces QColor instances with Color, creating two new global variables in `vibra/__init__.py`
- Updates renderer methods to accept Color objects instead of raw tuples, adjusting the logic necessary.


andrefpf (approved):
Everything looks fine, and the code already looks much better. Good job.

PR #154, opened 2025-07-18
Mesh quality control tools

vitorslongo:
This Pull Request adds a mesh quality control feature, which evaluates mesh quality based on several metrics: Gamma, Volume, Minimum Scaled Jacobian, and Aspect Ratio.

It computes and displays key statistics (worst value, mean, and standard deviation) for each metric. Users can also plot histograms and visually isolate low-quality elements in the render.

This feature is essential, as poorly shaped elements can cause numerical instabilities and lead to inaccurate analysis results.


vitorslongo:
KNOWN ISSUES, need help with these @andrefpf, please.
- the quality table on the mesh quality tab doesnt change the second time we generate the mesh, but works all the other times. Maybe it has something to do with the `None` atribution to some variable that's messing everything up
- the issue above also affects the histograms and the bad elements plotting in the render
- for the attached geometry the bad elements plot doesnt work for some reason, there are issues with the main window, mesh render etc. **Big problem**, help needed on this one André [geometria_colapsada.zip](https://github.com/user-attachments/files/21378965/geometria_colapsada.zip)

Also, for the PR to be promoted:
- the merge conflicts must be solved 
- new example files with the new mesh quality data must be created and saved to the examples folder
- the histograms plots need to be embedded in a QtWidget
- if it's easy we can also implement the stop if some collapsed element is detected, but maybe this one can be done in another PR/Issue, this PR is already huge as of now. Can you shed some light here André?

PR #168, opened 2025-07-31
Fix bugs detected during the import of table data in the structural model

JacsonGilVargas:
This pull request addresses issues detected during the import of table data for structural boundary conditions or dofs prescription. It also corrects bugs raised during the processing of the structural boundary condition symbols.


vitorslongo (commented):
image

There's a button with the autodefault setting activated where I think it shouldn't be

vitorslongo (changes requested):
image

This button is set as autodefault where I think it shouldn't

vitorslongo (changes requested):
maybe we could try to convert the data to numbers to make it easier for the user?
I've tried to use a csv file with strings instead of ready-to-use numbers and got this (unhandled?) error 
image

PR #170, opened 2025-08-06
Data importer implementation

GuilhermeSPierri:
This pull request adds the `DataImporter` and `ImportedData` classes to handle file imports. The first class uses only static methods, which read files and, when necessary, strip out unnecessary headers.

The second class is a `@dataclass`, which stores metadata about each imported file.

Additional adjustments were made to the boundary condition interfaces.


JacsonGilVargas (approved):
I performed several tests importing text and spreadsheet files to configure structural and acoustic models, and everything worked as expected. The import of tabular data is now more reliable, even when headers span multiple lines. Great teamwork @R0drigoSchwartz and @GuilhermeSPierri.

PR #174, opened 2025-08-09
Remove fileboxes usage from ProjectFile

gcmartins:
- Filebox usage was removed from ProjectFile class.
- The unused TEMP_PROJECT_FILE global var was also removed.

The new workflow for project file now is:
   * On project saving: Archive all files in the TEMP_PROJECT_DIR folder to a `.vibra` file
   * On project loading: Extract all `.vibra` files into TEMP_PROJECT_DIR folder.


JacsonGilVargas (changes requested):
I've run some tests and project file management seems to be working as expected, while maintaining backward compatibility with files saved using fileboxes. The time spent saving project files has increased slightly, so we'll need to conduct more tests on large models. This is not a problem, but a consequence of the new approach to handling the project data files. In my tests, I have noticed that if I solve one acoustic harmonic problem first, then 'remesh' with a different element size and try to solve again, the solution processing does not appear to be occurring correctly. The speed at which the solution progress bar fills indicates that the solution itself is not running. We must ensure that each frequency step is processed after any mesh updating.

vibra/project_files/project_file.py
```diff
@@ -670,51 +629,56 @@ def delete_harmonic_solution(self):
...
+        if self.results_data_filepath.exists():
+            os.remove(self.results_data_filepath)
+
+    def archive_project(self, zip_path: Path):
+        with zipfile.ZipFile(zip_path, 'w', zipfile.ZIP_DEFLATED) as zipf:
+            for root, dirs, files in os.walk(self.path):
+                for file in files:
+                    file_path = os.path.join(root, file)
+                    arcname = os.path.relpath(file_path, self.path)
+                    zipf.write(file_path, arcname)
```
andrefpf, line 677:
```suggestion
        for path in self.path.rglob("*"):
            if path.is_file():
                arcname = path.relative_to(self.path)
                zipf.write(path, arcname)
```

vibra/interface/project/save_project_data_selector.py
```diff
@@ -76,5 +78,15 @@ def proceed_callback(self):
...
+    
+
+def get_folder_size(path: Path):
+    total = 0
+    for dirpath, dirnames, filenames in os.walk(path):
+        for f in filenames:
+            fp = os.path.join(dirpath, f)
+            if os.path.isfile(fp):
+                total += os.path.getsize(fp)
+    return total
```
andrefpf, line 90:
```suggestion
def get_folder_size(path: Path):
    return sum(f.stat().st_size for f in path.rglob("*") if f.is_file())
```
Not sure if it sacrifices too much the legibility, it is a nice option

andrefpf (changes requested):
Besides the small suggestions related to the code, it seems like the solution values are not being cleaned when we we modify the mesh and re-run the analysis.

image

Also, I was not able to run a structural analysis. Not sure If something was not well configured or if it is actually an error. 

My setup consists in adding a x=1 nodal load on surface 3 of the default cylinder and running a structural harmonic analysis in the range(5, 600, 50)

image

gcmartins:
I've created a button below for properly resume the analysis when it is possible (I mean, when harmonic solution has stopped in the middle):
image

This button is only enabled when the model or the mesh is not changed when the project has being restored. 

I rapidly choose this icon, so feel free to suggest another if you find it better.

andrefpf (approved):
Now everything seems to be working properly.

You created a button to resume the solution, and a dialog that appears when the software loads a incomplete solution (as shown in the figure bellow).

I think that instead of this, we could have a single dialog that appears when the "run" button is clicked and it is possible to resume the solution. This dialog would ask if the user wants to run a whole new analysis, resume the previous analysis or cancel.

What do you think? It is just an idea, fell free to dismiss if it does not make sense.

image

JacsonGilVargas (approved):
I've performed several tests and everything worked as expected. I forced the code execution interruption while running the solution, and all previously solved frequency steps were recovered after reopening the application. Excellent job, @gcmartins.

PR #177, opened 2025-08-14
Advances in structural model setup

JacsonGilVargas:
This pull request implements a series of enhancements during the structural model setup. The interface controls and inputs are dynamically adjusted to guide the user, avoiding the activation of 2D and 3D element formulations at the same time. Beyond that, if collapsed elements are detected during mesh post-processing, the "Mesh Configuration" item will be highlighted in red. The model solution will be disabled until the mesh-related issue is resolved. This pull request also fixes some bugs encountered while processing the particle velocity plot after loading already solved model data.


vitorslongo (approved):
Everything seems to be working as expected, this changes make running the analises much simpler as well

gcmartins:
I think this is a very good workflow direction. But it still need some polishing on the app control and feedback when user mixes setups from structural and acoustic models, like: 

image

But it can be done afterwards, this PR is already too big.

Naedlig (commented):
It’s good to hear that the improvements have made the analyses smoother and the workflow more intuitive. Furthermore, it fixes an issue when importing external files. Therefore, everything appears to be working correctly. Congratulations, Jacson, on your excellent work.

PR #181, opened 2025-08-18
Reduce the number of animation cache frames

andrefpf:
This PR modifies the phases that are presented in a way that only half of the values need to be cached. A bonus of this approach is that all analysis appear by default with the phase that causes the maximum amplitude.


JacsonGilVargas (approved):
Now the animations are working as expected @andrefpf and @vitorslongo.

PR #184, opened 2025-08-26
Refactor linear solver to handle symmetric and complex inputs

### LINKED ISSUE #176: Decouple harmonic and modal analysis from physical models
Currently we have special analysis classes for each physical model. Many logic of these classes could be on the model or assembly classes. Therefore, we can have just one Harmonic and Modal analysis that receives a model to perform that analysis on it.
This could simplify and reduce logic duplication on these analysis classes.

gcmartins:
This is part of decoupling harmonic and modal solver efforts. Other PRs will be open to handle other parts of it.
This should work the same way as before and can be merged independently of the other parts of the work.


JacsonGilVargas (changes requested):
I've performed a few tests with acoustic analyses and found a problem while accessing the complex acoustic modal analysis data. The same problem doesn't appear in the non-complex acoustic modal analysis.


image


The structural harmonic analyses is showing inconsistent results, even for undamped problems.


image


What we should expect as a consistent result for the tested model at 325Hz, for instance.


image


Finally, structural modal analysis also yields physically inconsistent results. Fortunately, the dev branch is working well and can be used as a reference for addressing issues related to global matrices.

gcmartins:
> I've performed a few tests with acoustic analyses and found a problem while accessing the complex acoustic modal analysis data. The same problem doesn't appear in the non-complex acoustic modal analysis.
> 
> image
> The structural harmonic analyses is showing inconsistent results, even for undamped problems.
> 
> image
> What we should expect as a consistent result for the tested model at 325Hz, for instance.
> 
> image
> Finally, structural modal analysis also yields physically inconsistent results. Fortunately, the dev branch is working well and can be used as a reference for addressing issues related to global matrices.

I've found two issues while investigating this problem:

- Since we're now detecting symmetric matrices, the structural matrix is ​​showing some strange non-symmetric entries, I don't know why. Even in a free-form condition, it still has non-symmetric entries. I think we should investigate this.
- This non-symmetric matrix shouldn't be the main cause of the strange result you found. PARDISO supports non-symmetric matrices, but I don't think it's working well in this case. If you switch to the MUMPS solver, the results will be practically the same as before.

The results were consistent before because we were assuming the matrices were symmetric instead of detecting them and passing the symmetric matrix flag to PARDISO.

gcmartins:
> I noticed that only real non-symmetric solver (PARDISO) is giving weird results, so I change the logic to enforce complex non-symmetric solver for any non-symmetric matrix.


I think the acoustic modal analysis issue is due to this same reason. The PARDISO real non-symmetric solver is giving weird results.

andrefpf (approved):
As long as I tested everything seems to be working fine.

JacsonGilVargas (approved):
I've conducted several tests in both modal and harmonic analysis in the acoustic and structural physical domains, and everything worked very well. The proposed refactorization reorganized the code, with great emphasis on the solvers. Excellent work @gcmartins, congratulations.

PR #190, opened 2025-09-03
New viscous thermal loss

R0drigoSchwartz:
This PR contains some improvements to the `ViscousThermalLossModelInputs` class, providing the functionality to modify the attributes of a Viscous Thermal Loss model that has already been applied.

Now the `RectangularDuctData` class is used to represent a Viscous Thermal Loss model in a rectangular duct, and `CircularDuctData` to represent a Viscous Thermal Loss model in a circular duct.

Furthermore, the group selection has been disabled to prevent problems.


JacsonGilVargas (approved):
I ran a couple of tests and everything works fine. The model mapping implementation allows for easy editing of the existing viscous-thermal loss models. Excellent feature, congratulations.

PR #192, opened 2025-09-08
Fix visualization for all types of elements in multimaterial actor

andrefpf:


JacsonGilVargas (approved):
The surface's appearance has been recovered. Excellent work.

PR #199, opened 2025-09-18
Reset structural DOF prescription by assignment

JacsonGilVargas:
This pull request adds the DOF prescription reset whenever there is an assignment of "All DOF free". It also fixes some bugs found when removing table files related to the structural boundary conditions.


vitorslongo (commented):
Everything seems to be working fine.
Just one thing I'd change is to make the "Attribute" button AutoDefault.

vini-ribeiro (approved):

PR #204, opened 2025-09-20
Implements a temporary stop button

JacsonGilVargas:
This pull request implements a simple stop button to control the code execution break during the model solution. Admittedly, this solution isn't the most elegant, but it is simple and works perfectly. Feel free, @andrefpf and @gcmartins, to improve the proposal.

### DIFF
--- a/vibra/engine/assemblers/acoustic_assembler.py
+++ b/vibra/engine/assemblers/acoustic_assembler.py
@@ -933,8 +933,14 @@ class AcousticAssembler:
         # global_matrices shape
         self.gm_shape = (self.total_dof, self.total_dof)
 
+        logging.info(f"Processing the elementary matrices data... [25/100]")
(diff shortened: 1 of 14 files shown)


gcmartins:
> To be honest I do not think this is a very reliable solution.
> 
> Adding a model specific function inside a generic class such as `LoadingWindow` looks like a bad idea. Also, using the string `"Solution step"` to control the process button is a workaround that maybe could be replaced by a boolean in the class constructor.
> 
> Despite that, I do not oppose myself to merging it in the current stage for the following reasons:
> 
> * It is a very useful feature, specially while working with big models
> * Everything seems to be working as intended for the current use case
> * Better alternatives will probably require more time to be developed since they might require multiprocessing handling to work for modal analysis as well

vitorslongo (approved):

gcmartins (approved):

vitorslongo (approved):
Everything works as expected👍

Issue #132, opened 2025-06-05
Magnification factor bar changes value when hovering the mouse


vini-ribeiro:
When doing an analysis, if I click on the line of the magnification bar (not on the blue circle), the value changes, and whenever I hover the mouse over the bar, it keeps changing (even without clicking). The behavior returns to normal when I click on the blue circle.
It feels as if the mouse grabs the blue circle and doesn't release it until I click on it again. 

https://github.com/user-attachments/assets/edef0c2f-e9d3-478b-a96f-59dcbbcac582

PR #205, opened 2025-09-22
Fixes behavior bug in magnification factor slider

R0drigoSchwartz:
This PR fixes the behavior bug of the magnification factor slider when the user runs a structural analysis animation. To do this, when the user selects a new magnification factor, the animation stops, if it's running.

This PR closes the #132 issue.


vini-ribeiro (approved):
Good job, @R0drigoSchwartz! 🚀

vibra/interface/animation_toolbar.py
```diff
@@ -242,7 +243,7 @@ def process_animation(self, state: bool):
             )
         else:
             app().main_window.results_widget.stop_animation()
-
+
```
andrefpf, line 246: :+1:

andrefpf (approved):
LGTM

PR #215, opened 2025-09-24
Geometry integration Part 1

### LINKED ISSUE #217: Refactor the software to use Geometry class

GuilhermeSPierri:
This is an initial pull request to integrate the `Geometry` class into `VIBRA`. The goal is to decouple all Geometry-related elements from the Mesh class.

The main changes were getting all the topological relations using Geometry methods, not accessing Mesh dicts anymore.

### **What's new?**

- Now we can get the bounding box coordinates of the solids, surfaces and curves.
-  Optimized the normal calculation for points and curves.
- One single dictionary for each type of mapping, not using bidict anymore.
- New Geometry tests.


andrefpf (changes requested):
I found a bug. Here are the steps to replicate:
- Open `rectangular_cavities.STEP`
- Generate a mesh with default values
- Add a DOF Decoupling on `Surface 6`
- **Close the window** (this part is very important)
- Remove the DOF Decoupling
- Hide `Volume 2`

After these steps the `Surface 6` is being hidden as well.

vibra/engine/geometry/geometry.py
```diff
@@ -299,64 +305,153 @@ def _process_geometry_information(self):
...
+    self._solids_to_surfaces = deepcopy(self._cache_solids_to_surfaces)
+    self._surfaces_to_curves = deepcopy(self._cache_surfaces_to_curves)
+    self._curves_to_points = deepcopy(self._cache_curves_to_points)
+
+    self._cache_solids_to_surfaces.clear()
+    self._cache_surfaces_to_curves.clear()
+    self._cache_curves_to_points.clear()
+
+    # To be continued
+
```
andrefpf: I think these methods are not indented correctly

vibra/interface/model_inputs/acoustic/internal_impedances/perforated_plate_model_inputs.py
```diff
@@ -735,7 +737,7 @@ def remove_all_line_properties_boundind_surface(self, new_surface_ids: list[int]
 
         line_properties = deepcopy(self.properties.line_properties)
         for new_surface_id in new_surface_ids:
-            lines_from_surface = self.mesh.lines_from_surface.get(new_surface_id)
+            lines_from_surface = self.geometry.surfaces_to_curves(new_surface_id)
             if lines_from_surface is None:
                 continue
```
andrefpf: Some of these checks are not needed anymore, since surfaces_to_curves will never return None.

vibra/interface/model_inputs/acoustic/internal_impedances/perforated_plate_model_inputs.py
```diff
@@ -688,7 +690,7 @@ def include_user_defined_transfer_impedance(self, model: PerforatedPlateData, su
 
     def decouple_degrees_of_freedom(self, surface_id: int):
 
-        volumes_from_surface = self.mesh.volumes_from_surface.get(surface_id)
+        volumes_from_surface = self.geometry.surfaces_to_solids(surface_id)
         if volumes_from_surface is None:
             return
```
andrefpf: Not needed anymore.

andrefpf (changes requested):
More comments...

andrefpf:
Sorry @GuilhermeSPierri. In another moment we can resolve most of this.

PR #222, opened 2025-10-05
Update the mesher controls based on geometry information

JacsonGilVargas:
This pull request updates the mesher setup controls based on the geometry information. The primary objective is to control the topology of the available elements when working with shell elements.


vitorslongo (approved):
It's a nice detail to add.
Everything working as expected as well👍👍

PR #223, opened 2025-10-07
Fix section plane during animation

andrefpf:
This solves the bug reported internally by @vitorslongo, which happened when a section plane was applied during an animation. Besides the actual problem solving, I also opted to stop the animation during the cut, because it seems a more natural behavior. 

Please check if the problem was correctly solved and if the behavior looks correct.


JacsonGilVargas (approved):
I've run a couple of tests and everything worked fine. Excellent work, @andrefpf.

PR #224, opened 2025-10-07
Update exec config

andrefpf:
Update small problems regarding the executable and installable generation.


R0drigoSchwartz (approved):
Nice job @andrefpf!

PR #228, opened 2025-10-07
Add vibra logo on all renderers

andrefpf:
This PR adds the vibra logo on all renderers (as was originally requested, but I misinterpreted).

Since I could not find the dark and light color of the expanded logo, I used the short one.
When we have the expanded logo this can be easily updated.


Naedlig (approved):
The vibra logo is positioned and it works correctly. Congratulations  @andrefpf.

PR #236, opened 2025-10-13
Improve performance of analysis plots 

andrefpf:
This PR improves the performance of the analysis plots (specially on animation cache) by tweaking the method used to display the colors.

Previously each node received its own 3 or 4 components color scalar in a `vtkUnsignedCharArray`.
This approach was a bit more flexible, since any combination of colors could be set, but it also hindered the performance, since we needed a **long and scary loop in python** to set the color for each node.

This refactoring changed the color scalar to be a `vtkFloatArray`, holding a single number per node. This value is then mapped automatically by vtk according to a colormap which was also configured into the actor mapper.
Numpy was then used to pass the array data into the actor data with smaller overhead.

### DIFF
--- a/vibra/interface/main_window.py
+++ b/vibra/interface/main_window.py
@@ -892,40 +892,49 @@ class MainWindow(MainWindow_UI):
         self.model_setup_widget.model_setup_items.hide_model_setup_top_items()
 
         try:
-            if project_path is not None:
-                project_path = Path(project_path)
-                app().config.add_recent_file(project_path)
-                app().config.write_last_folder_path_in_file("project_folder", project_path)
-                self.update_recents_menu()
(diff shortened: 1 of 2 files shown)


JacsonGilVargas (approved):
Great work! After running a series of tests on both domains, I could verify the performance increase in processing and animation execution. The time spent on data post-processing was also considerably reduced. This pull request implements, without a doubt, the biggest efficiency gain in renderers I've seen in the app so far. My sincere congratulations, @andrefpf, on the excellent work.

Naedlig (approved):
The adjustments made have significantly reduced animation time. And they're working very well. Great jobs @andrefpf !

vitorslongo (approved):
Everything working as expected👍👍

PR #241, opened 2025-10-14
Delete properties with key press

### LINKED ISSUE #269: Delete properties by pressing delete on the keyboard
Specially when there are more than one property in a entity it may be a bit annoying to search the entries
needed to remove the attributes one by one.

A more convenient solution would be to delete the undesired properties by selecting the entity and pressing `delete`.

vini-ribeiro:
This PR adds a shortcut for the Delete (Del) key that allows removing acoustic or structural properties without having to manually remove them in each property widget. When the user selects one or more volumes, surfaces, lines, or points, and presses the Del key, a QDialog appears to let the user choose which property to delete. The selected property is then removed from all selected entities simultaneously.


JacsonGilVargas (approved):
I ran some tests applying and removing different properties, and the property removal feature worked perfectly. From what I could test, removing tabular data also worked as expected. Congratulations, @vini-ribeiro. Excellent feature.

andrefpf:
When we only have a single property in the selection and press `delete` the whole window appears.
Is it really necessary? Maybe when there is only one thing to remove we could bypass the selection phase and only remove the thing.

@JacsonGilVargas what is your take on it? It could also can be implemented in another moment.

image

andrefpf (changes requested):
When we run a solution and then remove some property using this method, the solution is maintained. The results should be removed whenever the model changes (adding or removing boundary condition) to ensure that the user will not get confused and associate a result with a non-correspondent model configuration.

This beavior is already correct when the property is removed by the conventional `List` method.

JacsonGilVargas:
I see no problem in proceeding with the removal of the property @andrefpf, as long as the user confirms the action. I spoke with other colleagues this afternoon and we decided to keep the user confirmation required to process the property removal. Regarding the solution reset after any changes to the model configuration, I believe we should discuss and align on the best strategy to resolve this issue in another PR. Removing properties through the respective property UI doesn't always reset the model results, so we may need to invest more time in this task in future updates.

andrefpf:
> I see no problem in proceeding with the removal of the property @andrefpf, as long as the user confirms the action. I spoke with other colleagues this afternoon and we decided to keep the user confirmation required to process the property removal. Regarding the solution reset after any changes to the model configuration, I believe we should discuss and align on the best strategy to resolve this issue in another PR. Removing properties doesn't always reset the model results, so we may need to invest more time in this task in future updates.

Oh. In this case I will approve the PR.

andrefpf (approved):

PR #254, opened 2025-10-20
Fix vibra logo

andrefpf:
I could not find a way using only PySide6 that could make the icon appear in windows.
So I used the function `SetCurrentProcessExplicitAppUserModelID` from `windll` and this aparently solved the issue.

@JacsonGilVargas I tested this on windows 11, but I guess you have access to a computer with windows 10. Please check if it works there as well.

@vini-ribeiro I am requesting your review to make sure these changes did not affected the Linux version.

### DIFF
--- a/vibra/__init__.py
+++ b/vibra/__init__.py
@@ -4,14 +4,16 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from vibra.interface.application import Application
 
-# copying the version from pyproject.toml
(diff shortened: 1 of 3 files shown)


JacsonGilVargas (approved):
I've tested the Vibra in 4 different computers with both W10 and W11 OS. The Vibra logo is now appearing normally, as it should be. Excellent, @andrefpf .

PR #259, opened 2025-10-23
Ignore refprop warnings

vitorslongo:
This PR adds proper handling for REFPROP warnings. Previously, Vibra treated them as errors, which prevented adding fluid configurations that triggered warnings. Now, they are handled correctly as warnings and only produce a warning message instead of an error.


JacsonGilVargas (approved):
I've conducted a few tests and everything seems to be working as expected. Excellent contribution, @vitorslongo.

PR #283, opened 2025-11-12
Implement cutter plane

### LINKED ISSUE #272: Test clipper function in vtk
@gcmartins found these two references that could help with the creation of a better section plane.
https://tutorial.pyvista.org/tutorial/08_widgets/e_plane-widget.html
https://examples.vtk.org/site/Python/VisualizationAlgorithms/Cutter/

Our hope is to have a "smooth" section plane (although it might also be useful to still keep the "whole elements" option).

Our current section plane is also very slow for detailed meshes.
Improving this performance would be great.

gcmartins:
https://github.com/user-attachments/assets/49df9efb-14e8-40de-ae37-bbc2ebf16a37

### DIFF
--- a/vibra/interface/viewer_3d/render_widgets/results_render_widget.py
+++ b/vibra/interface/viewer_3d/render_widgets/results_render_widget.py
@@ -34,6 +34,7 @@ class ResultsRenderWidget(AnimatedRenderWidget):
 
         app().main_window.theme_changed.connect(self.update_theme)
         app().main_window.section_plane.value_changed.connect(self.update_section_plane)
+        app().main_window.section_plane.value_changed.connect(self.update_color_and_deformation)
         app().main_window.visualization_changed.connect(self.visualization_changed_callback)
 
         self.acoustic_post = app().project.acoustic_postprocessing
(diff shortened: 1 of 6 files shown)


andrefpf (changes requested):
This functionality looks great, as well as the code.

I just found a minor bug with the colors on the first time when the mesh is sliced using the "jagged" way.
It looks like the color is not being set to both mappers. 

This problem is probably because I used `self.GetMapper()` in some places, and this function is returning only the currently used mapper, therefore a update on one of them is missing now.

image

image

Another thing that looks slightly weird are the lines passing in front of the mesh.
This could be solved using this dirty little trick to prioritize the mesh on the zbuffer, but since it admittedly very hacky, so it is up to you to use it or not.

```python
self.cutter_mapper.SetResolveCoincidentTopologyToPolygonOffset()
self.cutter_mapper.SetRelativeCoincidentTopologyPolygonOffsetParameters(-n, -m)
```

image

gcmartins:
I think now it is fixed:

https://github.com/user-attachments/assets/2f20e018-49a9-4df3-ba0e-d4fe0d63c7f3

vibra/interface/viewer_3d/render_widgets/results_render_widget.py
```diff
@@ -464,14 +465,21 @@ def _apply_section_plane(self, position, rotation, inverted, show_plane=True, is
             normal = -normal
 
         logging.info("Applying cut... [70/100]")
+        visualization = app().main_window.visualization_filter
         if is_mesh_field:
             self.analysis_actor.apply_cut(xyz, normal)
+            self.edges_actor.VisibilityOn()
+            visualization.lines = True
+            app().main_window.action_line_view.setChecked(True)
```
gcmartins: I'll try to refactor it later. This is really horrible.

andrefpf (changes requested):
The same problem is still happening when cutting in the mesh in the mesh workspace.

It is getting either red or blue when cut for the first time.

gcmartins:
> The same problem is still happening when cutting in the mesh in the mesh workspace.
> 
> It is getting either red or blue when cut for the first time.


Good catch, fixed now.

andrefpf (approved):

PR #287, opened 2025-12-04
Fix Zoom to Fit issue caused by symbols

vini-ribeiro:
The "Zoom to Fit" feature was failing to calculate bounds correctly when symbols were present. This fix forces a state refresh by temporarily deactivating and reactivating the symbols, ensuring the camera resets properly.


andrefpf (changes requested):
This seems to work, but maybe it is not the best solution.

VTK actors have this method to disable the bounds calculation.
```
actor.UseBoundsOff()
```

As far as I remember, if the bounds are disabled they do not affect the camera reseting, since these bounds are used to calculate the space occupied by the actors on the screen.

Please, try to use this method instead.

Issue #244, opened 2025-10-16
Create a class to handle selection


andrefpf:
The selection handling is currently controlled by a bunch of sets and functions dumped on the main window. 
It is already complicated enough to be handled in a dedicated class just for this. 

We also may add different selection modes, such as filters to control what type of mesh or geometry should be selected.

PR #290, opened 2025-12-11
Create a SelectionHandler class

GuilhermeSPierri:
This pull request moves all selection logic related to Geometry and Mesh elements into a dedicated class: `SelectionHandler`.

As part of the refactoring process, the goal of this PR is to decouple all the selection features from `MainWindow`, delegating this responsibility to the new class.

This PR addresses the initial part of issue #244. The next step is to add different selection modes.


vibra/interface/general/selection_handler.py
```diff
@@ -0,0 +1,120 @@
+from PySide6.QtCore import Signal, QObject
+from vibra import app
```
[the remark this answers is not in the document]
GuilhermeSPierri: > Could we avoid using app here? At first sight, we could include `project` or `model` as parameter on instantiation and get rid of it.

I agree that we should avoid using `app` inside `SelectionHandler`. Injecting the required dependency (in this case, `Mesh`) via the constructor would be a better approach and reduce coupling.

vibra/interface/general/selection_handler.py
```diff
@@ -0,0 +1,122 @@
...
+
+        if volumes is None:
+            volumes = set()
+
+        surfaces = set(surfaces) - set(self.hidden_surfaces)
+        volumes = set(volumes) - set(self.hidden_volumes)
+
+        # Select the surfaces associated to the selected volumes
+        for volume in volumes:
+            volume_surfaces = self.mesh.surfaces_from_volume.get(volume, [])
```
andrefpf: The type for `self.mesh` in this case is `Mesh | None`, therefore you need to check if `self.mesh is not None` before using it to avoid errors.

andrefpf (changes requested):
Unfortunately I found a little bug regarding the selection of volumes.

vibra/interface/general/selection_handler.py
```diff
@@ -0,0 +1,122 @@
+from typing import TYPE_CHECKING
+from PySide6.QtCore import Signal, QObject
+
+if TYPE_CHECKING:
+    from vibra.engine.mesher.mesh import Mesh
+
+class SelectionHandler(QObject):
+    selection_changed = Signal()
+    
+    def __init__(self, mesh: "Mesh | None" = None):
```
andrefpf:
```suggestion
    def __init__(self, mesh: Mesh | None = None):
```

It works without quotes if you add `from __future__ import annotations` on the imports.
Both options are a bit ugly, but importing from future improves the code highlighting (at least in my editor).

vibra/interface/general/selection_handler.py
GuilhermeSPierri: > If you create a project with a new geometry, like `new > cylinder.step` and select a face while holding shift (which is a hidden trick to select volumes), a bug appears.
> 
> It is probably because `selection.mesh` is not being update while loading geometries/meshes.

Yes, you were right. This bug was occurring exactly because of that.
The appropriate solution was to replace the use of `Mesh` with `Project`, since this instance is never outdated.

gcmartins (approved):

andrefpf (approved):
Well done!

PR #291, opened 2025-12-16
Migrate poetry to uv

R0drigoSchwartz:
This PR migrates the project from `Poetry` to `uv` as the dependency and project manager.

The migration aims to improve performance, since `uv` provides significantly faster dependency resolution and installation, while maintaining full compatibility with modern Python project workflows.

It is also a lot more ergonomic and automatically updates the environment whenever needed.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.10, <3.14"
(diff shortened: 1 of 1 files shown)


andrefpf:
Yeah, the CI/CD passed first try.

andrefpf (approved):

PR #297, opened 2026-01-26
Replace pandas with polars

R0drigoSchwartz:
This PR contains the replacement of Pandas with Polars.

Polars is library for data manipulation just like Pandas, but it's much faster, because it's written in Rust language.

Pandas library is maintained because we're using it's `ExcelWritter` class to save `.xlsx` files, since it has a better display format.


JacsonGilVargas (approved):
I performed a series of tests, and everything worked perfectly. The time spent importing the libraries needed to load the spreadsheet data was considerably reduced. Excellent job, @R0drigoSchwartz.

PR #312, opened 2026-02-10
Flexibilize the solution steps setup in harmonic analysis

### LINKED ISSUE #300: Filter DC component in acoustic analysis
To avoid mathematical inconsistencies while solving acoustic models, it's important to ensure that no zero-frequency component is present in the analysis setup. Some boundary conditions still allow the input of the zero-frequency component when importing tabular data. We should implement a filter to ignore this component just after importing tabular data.

### LINKED ISSUE #307: User-defined frequencies control
It would be interesting to have the ability to control the solution steps in harmonic analysis. Although it's not a recommended practice, it may be useful when solving large coupled models.

### LINKED ISSUE #313: Export natural frequencies to a table/file.
Export natural frequencies to a table/file.

JacsonGilVargas:
This pull request includes updates that make choosing solution steps more flexible in harmonic analysis, allowing the user control which step solution will be solved. Now, it's possible to change frequency boundaries and steps of tabular data. In addition, a set of user-defined solution steps can be selected from tabular data or manually entered by the user. A zero-frequency filter was also applied during the import of tabular data related to acoustic models to avoid mathematical inconsistencies in the resolution of these models. Finally, a new feature was implemented to export the list of frequencies and, when applicable, the damping ratio obtained with modal analysis.


vibra/engine/assemblers/acoustic_assembler.py
```diff
@@ -88,24 +88,27 @@ def get_prescribed_dof_values(self):
 
         for key, data in self.properties.surface_properties.items():
             property, surface_id = key
-            if property == "acoustic_pressure":
+            if property != "acoustic_pressure":
+                continue
```
andrefpf, line 92: This pattern makes me happy

andrefpf (commented):
I think it is a nice touch to keep the frequencies from the already configured uniform approach.
Image

vibra/interface/analysis/harmonic_analysis_setup_input.py
```diff
@@ -88,21 +136,57 @@ def analysis_method_callback(self):
...
         f_step = self.analysis_setup.get("f_step", 5)
-        global_damping = self.analysis_setup.get("global_damping", (0, 0, 0))
+        global_damping = self.analysis_setup.get("global_damping", (0., 0., 0.))
 
         self.load_analysis_type()
         self.load_damping_inputs(self.analysis_id, global_damping)
+        self.update_solution_steps_controls_visibility()
         self.load_frequency_setup_inputs(f_min, f_max, f_step)
 
+        if self.analysis_setup.get("frequency_spacing") == "user-defined":
```
andrefpf: We already talked about how writing a string in the file is better than to use a numeric enum.
But I just found that we can have StrEnum to use the best of both words:
- Well delimited universe of possibilities
- Easy JSON handling (I guess)
- Easy to read in the file

The code should look like this:
```python
from enum import StrEnum

class FrequencySpacingTypes(StrEnum):
    USER_DEFINED = "user-defined"
    EQUALLY_DISTRIBUTED = "equally distributed"
```
This is not strictly needed, but you may also find it useful.

vibra/interface/analysis/harmonic_analysis_setup_input.py
```diff
@@ -138,18 +223,53 @@ def load_frequency_setup_inputs(self, f_min: float, f_max: float, f_step: float)
...
+        f_min_tab, f_max_tab, *_ = self.tabular_frequency_setup
+    
+        if f_min < f_min_tab:
+            if not np.isclose(f_min, f_min_tab, 1e-8):
+                self.hide()
+                title = "Invalid minimum frequency"
+                message = "The value entered for the minimum frequency is out of the allowable range."
+                PrintMessageInput([error_title, title, message])
+                self.lineEdit_fmin.setFocus()
+                self.lineEdit_fmin.setStyleSheet("border-color: rgb(255,0,0); border-width: 2px")
```
andrefpf: I still did not see it in practice, since I coudn't use a table yet, but I like the idea.
@R0drigoSchwartz wasn't there a qml option that automatically marked errors like this?

andrefpf:
Although we can select multiple frequencies, we can not remove them in bulk
image

Naedlig (approved):
The new features in Vibra that permits the user the control of frequency steps solution and export the modal analysis results are working correctly. Good job @JacsonGilVargas.

andrefpf (approved):
LGTM

vitorslongo (approved):
Everything seems to be working nicely 👍

olavosilva (approved):
Good work!

PR #315, opened 2026-02-13
Add selection menu

### LINKED ISSUE #306: Add selection menu
Now that we have a proper selection handle, we can improve the usability of the software by adding a Selection menu, between Settings and View.

Image

This menu must have the following buttons:
- Geometry select all
    - Points
    - Faces
    - Solids
    - All entities
- Mesh select all
    - Nodes
    - Surface elements
    - Solid elements
- Invert selection
- Clear selection
- Volume selection mode (This option must be checkable)

I think that the inverted selection must only invert a set if there is something selected in it.
For example, if you only have a single surface selected, and no points, lines or volumes, the inverted set must contain all surfaces, except the previously selected one.

Technically it should contain all points, lines and volumes as well, since the inverse of an empty set is the whole set.
But this behavior would look very weird for the user.

taianabsfarias:
A selection menu has been added to the interface, allowing users to quickly select, clear, and invert geometry and mesh entities, in addition to the volume selection mode.
Captura de tela 2026-02-13 170137
Closes #306.


vibra/interface/main_window.py
```diff
@@ -357,6 +357,99 @@ def action_user_preferences_callback(self):
         self.close_dialogs()
         self.render_user_preferences = RendererUserPreferencesInput()
 
+    def action_points_callback(self):
+        all_ids = app().project.model.mesh.all_point_ids()
+        self.selection.set_geometry_selection(points=all_ids)
+        self.update_plots()
```
andrefpf: This `self.update_plots()` is an expensive function and is not necessary for these operations.
The color is automatically updated whenever `SelectionHandler.selection_changed` is fired.

vibra/interface/main_window.py
```diff
@@ -357,6 +357,99 @@ def action_user_preferences_callback(self):
...
+            if selection.mesh_nodes:
+                    new_mesh_selection["nodes"] = mesh.all_node_ids() - selection.mesh_nodes
+            if selection.mesh_faces:
+                    new_mesh_selection["faces"] = mesh.all_face_element_ids() - selection.mesh_faces
+            if selection.mesh_solids:
+                    new_mesh_selection["solids"] = mesh.all_solid_element_ids() - selection.mesh_solids
+            if new_mesh_selection:
+                selection.set_mesh_selection(**new_mesh_selection)
+
+        self.update_plots()
```
andrefpf: I sugest moving this implementaion to inside SelectionHandler to reduce the complexity of the main_window code.
This callback can be turned into something like this:
```python
def action_invert_selection_callback(self):
    self.selection.invert_selection()
```

vibra/interface/main_window.py
```diff
@@ -357,6 +357,99 @@ def action_user_preferences_callback(self):
...
+            selection.geometry_points,
+            selection.geometry_lines,
+            selection.geometry_surfaces,
+            selection.geometry_volumes,
+        ])
+        mesh_selection_active = any([
+            selection.mesh_nodes,
+            selection.mesh_faces,
+            selection.mesh_solids,
+        ])
```
andrefpf: nice

andrefpf:
I just noticed that I forgot to include Geometry->Lines and Mesh->All entities in the issue description. 
Can you also add these options, please?

image

image

andrefpf (changes requested):
Everything looks pretty good, but I added a few minor requests to improve the readability and code performance.

andrefpf:
I noticed that this branch has conflicts with dev. 
These resources files are automatically generated and usually cause conflicts.

Fortunatelly they are very easy to solve. You just need to:
- Update your local dev branch with the latest changes
- Merge dev into your branch
- Recompile the UIs with `uv run inv ui-compile`

andrefpf (approved):
Great job, everything working as intended.

PR #323, opened 2026-02-24
Fix DataImporter

R0drigoSchwartz:
This pull request contains a small adjustment to verify the validity of file lines when it is imported using the DataImporter class. 

It fixes a bug that occurred when a file with multiple lines of text was imported.


JacsonGilVargas (approved):
Excellent, @R0drigoSchwartz. It's now working as expected.

Issue #318, opened 2026-02-19
Store file hashes in the project file


andrefpf:
Currently we micromanage every file writing, and it is a very error prone operation.
Unfortunatelly we can not automate many stuff, because rewriting some files (such as tables and meshes) can be very slow.

Hashing arrays, for example, is very useful because we can rewrite data only if it was modified.

To avoid unecessary complication, we can only hash the expensive files, and accept to rewrite the smaller ones.
Apparently [xxhash](https://pypi.org/project/xxhash/) is a very fast, non-cryptographic hash algorithm, that may fulfill our needs.

Issue #303, opened 2026-01-29
Create classes to store solutions


andrefpf:
The solution of a model is more than a simple array, it must have more associated information.
In the `ModalSolver` class there is a `solution` variable and a `natural_frequencies` variable, even though both of them are part of the solution.

Many useful information can also be added to the solution class, such as the name of the analysis, the analysis setup, min, max and mean values, hashes of the mesh, hash of model properties, and many more.
Not all of this information is needed, but a few of them may be useful to start.

Comprehensive methods to retrieve solution related data is also desirable.
It can be easily implemented using AcousticPostprocessing and StructuralPostprocessing.

LazyHDF5MatrixWriter already implements a few of these features, but since it is used exclusively with HDF5 files, the interface can not be used with in memory solutions.
This new solution class can be a good interface to thread lazy and non-lazy data without exposing too much the API user to inner implementation details.

andrefpf:
I sugest implementing it like this:

```python
class Solution:
    analysis_id: AnalysisId
    ...

class AcousticHarmonicSolution(Solution):
   analysis_id = AnalysisID.ACOUSTIC_HARMONIC

    def __init__(self):
        ...

class StructuralModalSolution:
    analysis_id = AnalysisID.STRUCTURAL_MODAL

    def __init__(self):
        ...
``` 

Multiple inheritance would make sense as well, but the code becomes disgusting really fast. 
Keeping the analysis_id as a class variable allows one to access the solution category as follows:

```python
solution.analysis_id.is_harmonic()
or
StructuralModalSolution.analysis_id.is_acoustic()
```

PR #324, opened 2026-02-25
Refactor Project class to decouple it from the interface

### LINKED ISSUE #277: Add the hability to load a project without the need of a GUI
The software was not planned to run a project without the interface.
It happens that it is a pretty useful feature, and it helps a lot with the execution on separate processes.

This solution requires a lot of refactoring, which can not affect the simulation results.

andrefpf:
This PR changes a lot of things, I am sorry for this. 
I know that many errors may arise, and I am expecting a bunch of reviewing comments, so please don't be shy.

Close #277 #318 #303


andrefpf:
Initially I just wanted to decouple Project from the interface, but since it was intended to run simulations by code, I decided to refactor many of the API parameters to make it more usable from an only-code perspective. Among the main problems of the previous API we can list the following:

- The configuration of many parameter was done by misterious dictionaries.
```python
mesh_setup = {"minimum_element_size": 50, "maximum_element_size": 50, "element_type": "tetrahedral", "shape_function": "linear"}
analysis_setup = {"analysis_id": 3, "f_min": f_min, "f_max": f_max, "f_step": df, "frequencies": frequencies}
```
- To do trivial things such as importing a geometry and generating a mesh, the user would need to call multiple functions, many of them a bit unexpected.
```python
...
model.set_geometry_path(path)
model.set_length_unit()  # yep, no parameter
model.set_geometry_quality_factor()  # another setter without parameter
model.initialize_mesh()  
model.set_mesh_setup(mesh_setup)
model.process_mesh()
```
- An important part of the vibra files are the material and fluid tables. Despite that, they were entirelly managed inside a QWidget classs, according to the user inputs.
- Since the working directory is a fixed path, two Project classes would conflict with each other.
- Two classes were used to manage the working directory files. One of them converted between the files and dictionaries, while the other one used the dictionaries to modify the classes in place (we could not only load specific parts of the project)

Since most of these problems were addressed, the whole codebased was adapted to use this new paradigm, that is why so many files were modified.

andrefpf:
At this point all uses of "app().project" were replaced with "app().new_project".

Since this already took a lot of time, I gave up with many refactors and aimed to "good enough".
This way I can work in other important demands in parallel, without the fear of having to deal with huge conflics.

vibra/engine/serialization/project_writer.py
```diff
@@ -0,0 +1,419 @@
...
+from pathlib import Path
+from typing import TYPE_CHECKING, Optional
+
+import h5py
+import numpy as np
+from PIL import Image
+
+if TYPE_CHECKING:
+    # This is to avoid circular imports since
+    # this file is imported by NewProject
```
gcmartins: I think this is a symptom of a lack of abstraction.
I think the current NewProject could be called ProjectManager and the NewProject class should handle mostly the project data.

andrefpf: That is fair enough. 

I am still confused about the data that should be managed though. 
`NewProject` already feels like a `ModelManager`.

Maybe I should move the few data variables to `Model` and make this an actual manager?
What do you think?

gcmartins: Yeah, it could be a good starting point. Model is another "God class" that we could break in the future. We could have a little weird abstraction on the Model for now and break it later. 
Actually, for the current model structure it is not that weird :smile:

vibra/engine/analysis_info/harmonic_analysis_setup.py
```diff
@@ -0,0 +1,62 @@
...
+    f_min: float
+    f_max: float
+    f_size: float
+    analysis_method: AnalysisMethod = AnalysisMethod.DIRECT
+    global_damping: tuple[float, float, float] = (0.0, 0.0, 0.0)
+    modes_number: None | int = None
+
+    def __init__(self, *args, **kwargs):
+        # Overriding ABC was not raising an error for me.
+        # Maybe there is a better way to prevent abstract initialization.
```
gcmartins: Just inheriting from ABC doesn't work here?
```
from abc import ABC

class HarmonicAnalysisSetup(ABC):
... 
```

vibra/utils/signal.py
```diff
@@ -0,0 +1,56 @@
+import functools
+
+
+class VibraSignal:
```
gcmartins: Is this really needed?
I'm not seeing any usage of it. It is also a bit weird to need it on scripting.

vibra/engine/serialization/_tests/test_mesh_serialization.py
```diff
@@ -0,0 +1,80 @@
...
+import numpy as np
+
+from vibra import PROJECT_DIR
+from vibra.engine.analysis_info import AnalysisID, HarmonicAnalysisSetupRange, ModalAnalysisSetup
+from vibra.engine.project import Project
+
+
+def test_write_and_read_mesh_project(fluid):
+    mesh_path = PROJECT_DIR / "data/examples/mesh_files/cavities_60mm_large.nas"
+    project_path = Path("project.vibra")
```
gcmartins: Let's use datadir fixture here to avoid running test on the current code path:
```suggestion
def test_write_and_read_mesh_project(fluid, datadir: Path):
    mesh_path = PROJECT_DIR / "data/examples/mesh_files/cavities_60mm_large.nas"
    project_path = datadir / "project.vibra"
```

vibra/engine/serialization/_tests/test_geometry_serialization.py
```diff
@@ -0,0 +1,53 @@
...
+
+from vibra import PROJECT_DIR
+from vibra.engine.analysis_info import AnalysisID, HarmonicAnalysisSetupRange
+from vibra.engine.mesher.mesh_setup import MeshSetup
+from vibra.engine.project import Project
+
+
+def test_write_and_read_project(fluid):
+    geometry_path = PROJECT_DIR / "data/examples/geometry_files/cylinder.step"
+    project_path = Path("project.vibra")
```
gcmartins:
```suggestion
def test_write_and_read_project(fluid, datadir: Path):
    geometry_path = PROJECT_DIR / "data/examples/geometry_files/cylinder.step"
    project_path = datadir / "project.vibra"
```

vibra/engine/serialization/_tests/test_mesh_serialization.py
gcmartins:
```suggestion
def test_write_and_read_mesh_project(fluid, datadir: Path):
    mesh_path = PROJECT_DIR / "data/examples/mesh_files/cavities_60mm_large.nas"
    project_path = datadir / "project.vibra"
```

vibra/engine/solvers/harmonic_solver.py
```diff
@@ -37,16 +39,27 @@ def solve_direct(self, print_log: bool = False, is_resume: bool = False):
...
+            return StructuralHarmonicSolution(
+                self.assembler.model.analysis_setup.frequencies(),
+                self.solution,
+                self.displacement_dof,
+            )
+        else:
+            return AcousticHarmonicSolution(
+                self.assembler.model.analysis_setup.frequencies(),
+                self.solution,
+            )
```
gcmartins: I'm not seeing much benefit on specialising the solution here.

vibra/engine/solution/structural_modal_solution.py
```diff
@@ -0,0 +1,46 @@
...
+
+import numpy as np
+
+from vibra.engine.analysis_info import AnalysisID
+
+from .common_solution import Array1D, Array2D, CommonModalSolution
+
+
+class StructuralModalSolution(CommonModalSolution):
+    analysis_id = AnalysisID.STRUCTURAL_MODAL
```
gcmartins: Is a bit ackward that the specialized solution still need the analysis_id.

gcmartins: I would say that better just have the Harmonic and Modal class of solution containers. Then we can dinamically mark the analysis_id in there. These specializations seems to be mostly duplicated codes.

vibra/engine/analysis_info/harmonic_analysis_setup.py
andrefpf: I tested ABC before adding any abstract method to the class, so it was being instantiated normally.
Now this message appears:
```python
>>> from vibra.engine.analysis_info.harmonic_analysis_setup import HarmonicAnalysisSetup
>>> a = HarmonicAnalysisSetup()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    import platform
        ^^^^^^^^^^^
TypeError: Can't instantiate abstract class HarmonicAnalysisSetup without an implementation for abstract methods 'as_dict', 'frequencies'
```

Although it is working now, I think that listing the implemented subclasses is a bit more useful in this case.
What do you think?

vibra/engine/solvers/harmonic_solver.py
andrefpf: Initially I was thinking of adding wrapper for some of the post processing results directly in the solution classes. But yeah, it is probably not a great idea.

vibra/engine/solution/structural_modal_solution.py
andrefpf: Good idea. I've implemented this way now.

vibra/engine/analysis_info/harmonic_analysis_setup.py
gcmartins: I prefer the ABC approach. It is basic but works.
IDK if it will ensure that the abstract methods are implemented in the subclasses.
What if the subclasses are moved/implemented in another module?

vibra/engine/solution/harmonic_solution.py
```diff
@@ -0,0 +1,82 @@
...
+
+            case None, _:
+                return False
+
+            case _, None:
+                return False
+
+            case _, _:
+                if not np.allclose(self.displacement_dof, other.displacement_dof):
+                    return False
```
gcmartins: I guess it works, doesn't it?
```suggestion
                return np.allclose(self.displacement_dof, other.displacement_dof):
```

vibra/engine/solution/harmonic_solution.py
```diff
@@ -0,0 +1,82 @@
...
+
+        match self.displacement_dof, other.displacement_dof:
+            case None, None:
+                pass
+
+            case None, _:
+                return False
+
+            case _, None:
+                return False
```
gcmartins: Do we really need it?

vibra/engine/solution/harmonic_solution.py
gcmartins: It seems to miss the other fields comparison, doesn't it?

andrefpf: It does not work. If the displacement dofs are different we can stop. If they are the same, we still need to compare the remaining fields.

vibra/engine/solution/harmonic_solution.py
andrefpf: Probably there is a better way, but we need to check if the types are the same in both classes.

vibra/engine/analysis_info/harmonic_analysis_setup.py
andrefpf: Actually it is not a problem, the module printing is being handled automatically.
But I will keep the ABC approach for simplicity.

vibra/engine/analysis_info/harmonic_analysis_setup.py
```diff
@@ -28,9 +28,13 @@ class HarmonicAnalysisSetup:
...
-        subclass_names = ", ".join([f"{cls.__module__}.{cls.__name__}" for cls in self.__class__.__subclasses__()])
+        super().__init__(*args, **kwargs)
+
+        subclass_names = [f"{cls.__module__}.{cls.__name__}" for cls in self.__class__.__subclasses__()]
         msg = "HarmonicAnalysisSetup can not be intantiated.\n"
-        msg += f"Use one of the following classes instead: {subclass_names}"
+        msg += "Use one of the following classes instead:\n"
+        for name in subclass_names:
+            msg += f"\t{name}\n"
         raise TypeError(msg)
```
gcmartins: I think these lines will never run.

gcmartins: I think this is overengineering something that is not that useful. Maybe better is make this class "private" to avoid scripting users to try to use it.

andrefpf: Removed the init

gcmartins:
Structural modal results is broken on showing mode shapes:
image

vibra/engine/solvers/modal_solver.py
```diff
@@ -96,7 +96,8 @@ def solve(self, which="LM", full_solution: bool = True) -> ModalSolution:
         else:
             cnf = None
 
-        self.displacement_dof = self.assembler.displacement_dof
+        if isinstance(self.assembler, StructuralAssembler):
+            self.displacement_dof = self.assembler.displacement_dof
```
gcmartins, line 103: Maybe better set a property method for it, instead of having placeholder for this attribute. Semantically, displacement_dof does not belong to the solution, it is just a shortcut.

vibra/engine/project.py
```diff
@@ -0,0 +1,449 @@
...
+    def get_analysis_type_and_physical_domain(self) -> tuple[str, str]:
+        analysis_type = ""
+        physical_domain = ""
+
+        if self.model.analysis_id.is_harmonic():
+            analysis_type = "harmonic"
+        elif self.model.analysis_id.is_modal():
+            analysis_type = "modal"
+        elif self.model.analysis_id.is_static():
+            physical_domain = "static"
```
gcmartins:
```suggestion
            analysis_type = "static"
```

vibra/engine/project.py
```diff
@@ -0,0 +1,449 @@
...
+
+        if self.model.analysis_id.is_acoustic():
+            physical_domain = "acoustic"
+        elif self.model.analysis_id.is_structural():
+            physical_domain = "structural"
+        elif self.model.analysis_id.is_coupled():
+            analysis_type = "coupled"
```
gcmartins:
```suggestion
            physical_domain = "coupled"
```

vibra/engine/project.py
```diff
@@ -0,0 +1,449 @@
...
+            return False
+
+        if self.solver.solution is None:
+            return False
+
+        return self.solver.solution.size > 0
+
+    def get_analysis_type_and_physical_domain(self) -> tuple[str, str]:
+        analysis_type = ""
+        physical_domain = ""
```
gcmartins: Why not using Enums here?

vibra/engine/model.py
```diff
@@ -85,7 +87,7 @@ def reset_variables(self):
         self.list_frequencies = list()
         self.solution_steps_mask = list()
```
gcmartins: I've created this proposal to try and improve this in the future: https://github.com/MOPT-UFSC/VIBRA/issues/334

andrefpf: Looks like a good workflow to me. It also seems a good starting point for that idea of using the FEM engine as an external library, separating it from the interface.

We could start a new repository building this as a proof of concept.

vibra/engine/solvers/modal_solver.py
andrefpf: I am not sure if I understood the suggestion here. To be fair I also don't understand the use case of this variable.

Do you mean to remove the variable from the solution class? Use as an property method of which class?

gcmartins: The displacement_dof is not used anywhere on the solver computing. I think it is here just as a shortcut to be used for post-processing.

vibra/engine/_tests/test_harmonic_acoustic.py
gcmartins, line 1: 🙏

vibra/engine/_tests/test_modal_acoustic.py
gcmartins, line 1: 🙏

validation_files/validate_acoustic_element_hex20_harmonic.py
```diff
@@ -187,15 +187,16 @@ def load_external_mesh_and_solve():
     # Run harmonic analysis
 
     t0 = time()
-    solution = harmonic_solver.solve_direct(print_log=True)
+    s = harmonic_solver.solve_direct(print_log=True)
```
R0drigoSchwartz: When attempting to perform the validation of the hex20 element, I received an error in the `solve_direct` method of the harmonic solver. The check to determine if the analysis setup is an instance of `HarmonicAnalysisSetup` fails because the analysis setup is `None`.

The error return function:
```
 File "C:\Users\Usuario\VIBRA\validation_files\validate_acoustic_element_hex20_harmonic.py", line 389, in <module>
    load_external_mesh_and_solve()
  File "C:\Users\Usuario\VIBRA\validation_files\validate_acoustic_element_hex20_harmonic.py", line 190, in load_external_mesh_and_solve
    s = harmonic_solver.solve_direct(print_log=True)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Usuario\VIBRA\vibra\engine\solvers\harmonic_solver.py", line 55, in solve_direct
    assert isinstance(analysis_setup, HarmonicAnalysisSetup)
AssertionError
```

The same error is happening for the other acoustic harmonic validations.

validation_files/validate_acoustic_element_hex20_harmonic.py
```diff
@@ -187,15 +187,16 @@ def load_external_mesh_and_solve():
...
+    solution = s.results
     dt = time() - t0
     print(f"Elapsed time to solve harmonic analysis: {round(dt, 4)}")
 
     input_rows = mesh.external_nodes_from_surfaces[1]
     output_rows = mesh.external_nodes_from_surfaces[2]
 
-    input_pressure = np.average(solution[input_rows, :], axis=0).flatten()
-    output_pressure = np.average(solution[output_rows, :], axis=0).flatten()
```
R0drigoSchwartz, line 198: Here you declare your `solution` variable as `s.results`:

```python
solution = s.results
```
That's supposed to be an `Array2D ` object. And then you do:

```python
input_pressure = np.average(solution.results[input_rows, :], axis=0).flatten()
output_pressure = np.average(solution.results[output_rows, :], axis=0).flatten()
```
Since the Array objects don't have the attribute results, it should be:
```python
input_pressure = np.average(solution[input_rows, :], axis=0).flatten()
output_pressure = np.average(solution[output_rows, :], axis=0).flatten()
```
Right?

validation_files/validate_acoustic_element_hex8_harmonic.py
```diff
@@ -187,7 +187,8 @@ def load_external_mesh_and_solve():
     # Run harmonic analysis
 
     t0 = time()
-    solution = harmonic_solver.solve_direct(print_log=True)
+    s = harmonic_solver.solve_direct(print_log=True)
```
R0drigoSchwartz: Same error here.

validation_files/validate_acoustic_element_tet10_harmonic.py
```diff
@@ -187,7 +187,8 @@ def load_external_mesh_and_solve():
     # Run harmonic analysis
 
     t0 = time()
-    solution = harmonic_solver.solve_direct(print_log=True)
+    s = harmonic_solver.solve_direct(print_log=True)
```
R0drigoSchwartz: Same here.

validation_files/validate_acoustic_element_tet4_harmonic.py
```diff
@@ -189,7 +189,8 @@ def load_external_mesh_and_solve():
     # Run harmonic analysis
 
     t0 = time()
-    solution = harmonic_solver.solve_direct(print_log=True)
+    s = harmonic_solver.solve_direct(print_log=True)
```
R0drigoSchwartz: Same.

validation_files/validate_mass_source_tet10.py
```diff
@@ -210,7 +210,8 @@ def load_external_mesh_and_solve(assignment_type: str):
     # Run harmonic analysis
 
     t0 = time()
-    solution = harmonic_solver.solve_direct(print_log=True)
+    s = harmonic_solver.solve_direct(print_log=True)
```
R0drigoSchwartz: Same.

validation_files/validate_mass_source_tet4.py
```diff
@@ -207,7 +207,8 @@ def load_external_mesh_and_solve(assignment_type: str):
     # Run harmonic analysis
 
     t0 = time()
-    solution = harmonic_solver.solve_direct(print_log=True)
+    s = harmonic_solver.solve_direct(print_log=True)
```
R0drigoSchwartz: Same.

validation_files/validate_porous_material_for_silencer.py
```diff
@@ -163,7 +163,8 @@ def load_external_mesh_and_solve():
     # Run harmonic analysis
 
     t0 = time()
-    solution = harmonic_solver.solve_direct(print_log=True)
+    s = harmonic_solver.solve_direct(print_log=True)
```
R0drigoSchwartz: Same.

validation_files/validate_shell_for_Lpipe.py
R0drigoSchwartz, line 1: Same problem here.

validation_files/validate_shell_for_piping_example.py
R0drigoSchwartz, line 1: Same here.

validation_files/validate_structural_element_hex8_harmonic.py
R0drigoSchwartz, line 1: In this file, the import line:
``` python
from vibra.engine.mesher.element_type import HEXAHEDRON_8
```

should be replaced by:
``` python
from vibra.engine.mesher.element_setup import HEXAHEDRON_8
```

validation_files/validate_structural_element_hex8_harmonic.py
```diff
@@ -168,7 +168,8 @@ def load_external_mesh_and_solve():
     t0 = time()
     # Run modal analysis
     harmonic_solver = HarmonicSolver(assembler)
-    solution = harmonic_solver.solve_direct(print_log=True)
+    s = harmonic_solver.solve_direct(print_log=True)
```
R0drigoSchwartz: Same.

validation_files/validate_structural_element_hex8_modal.py
R0drigoSchwartz, line 1: In this file, the import line:
``` python
from vibra.engine.mesher.element_type import HEXAHEDRON_8
```

should be replaced by:
``` python
from vibra.engine.mesher.element_setup import HEXAHEDRON_8
```

validation_files/validate_transmission_loss_for_complex_impedances.py
```diff
@@ -189,7 +189,8 @@ def load_external_mesh_and_solve():
     # Run harmonic analysis
 
     t0 = time()
-    solution = harmonic_solver.solve_direct(print_log=True)
+    s = harmonic_solver.solve_direct(print_log=True)
```
R0drigoSchwartz: Same.

validation_files/validate_viscous_thermal_loss_models.py
```diff
@@ -155,7 +155,8 @@ def load_external_mesh_and_solve():
     # Run harmonic analysis
 
     t0 = time()
-    solution = harmonic_solver.solve_direct(print_log=True)
+    s = harmonic_solver.solve_direct(print_log=True)
```
R0drigoSchwartz: Same.

vibra/engine/properties/libraries/property_library.py
```diff
@@ -0,0 +1,174 @@
...
+    def find_by_name(self, name: str) -> Optional[T]:
+        return self._find_by_attribute("name", name)
+
+    def contains_id(self, identifier: int) -> bool:
+        return identifier in self._data.keys()
+
+    def contains(self, obj: T) -> bool:
+        return obj in self._data.values()
+
+    def get_dupplicated_name(self, name: str) -> str:
```
vitorslongo:
```suggestion
    def get_duplicated_name(self, name: str) -> str:
```

vibra/engine/_tests/test_harmonic_acoustic.py
```diff
@@ -0,0 +1,78 @@
...
+    project.model.properties._set_property(
+        "specific_impedance",
+        {
+            "anechoic_termination": True,
+            "volume_id": 1,
+        },
+        surface=5,
+    )
+
+    solution = project.run_analysis()
```
R0drigoSchwartz, line 88: The  `run_analysis` method calls the `solver_direct` method, that generates the error previously mentioned.

vibra/engine/_tests/test_harmonic_structural.py
```diff
@@ -0,0 +1,74 @@
...
+            "element_type": "3d_element",
+            "real_values": [1, 0, 2],
+            "imag_values": [0, 0, 0],
+            "nodal_attribution": True,
+            "averaged": True,
+        },
+        surface=5,
+    )
+
+    solution = project.run_analysis()
```
R0drigoSchwartz, line 81: Same.

vibra/engine/serialization/_tests/test_geometry_serialization.py
```diff
@@ -0,0 +1,51 @@
...
+
+    project_a.configure_analysis(
+        AnalysisID.ACOUSTIC_HARMONIC,
+        HarmonicAnalysisSetupRange(
+            f_min=100,
+            f_max=500,
+            f_step=200,
+        ),
+    )
+    solution_a = project_a.solve_acoustic_harmonic_analysis()
```
R0drigoSchwartz, line 51: Same error happens here.

vibra/engine/serialization/_tests/test_mesh_serialization.py
```diff
@@ -0,0 +1,80 @@
...
+    project_cli.configure_analysis(
+        AnalysisID.ACOUSTIC_HARMONIC,
+        HarmonicAnalysisSetupRange(
+            f_min=200,
+            f_max=500,
+            f_step=100,
+            analysis_method="direct",
+        ),
+    )
+    project_cli.run_analysis()
```
R0drigoSchwartz, line 84: And here too.

R0drigoSchwartz (dismissed):
What an incredible job @andrefpf. Well done!

R0drigoSchwartz (commented):
Nice job @andrefpf , this is a great effort and significant improvement to the Vibra code.

On my review, only the tests I mentioned didn't work. Aside from that, everything seems great!

I will just wait you to modify the broken tests, and after this, I will accept this pull request.

Again, impressive job! 🚀

vibra/engine/properties/libraries/property_library.py
andrefpf: it seems like I duplicated the P letter

JacsonGilVargas (changes requested):
I've got the following error after pressing the button to export the acoustic transfer element data.

Image

<br> I think that converting the Path returned from the 'get_last_folder_for' method to a string might solve this issue.

JacsonGilVargas:
This error has occurred while trying to solve the acoustic model to export the element transfer data.

image

JacsonGilVargas:
The user is being asked to save the project data, even if the project has not been modified since it was opened.

image

JacsonGilVargas:
We can try to solve models without a mesh, which should never be possible.

image


<br>I strongly recommend bringing all already implemented checkers in dev to avoid this problem and its consequences. There are specific processes that require a valid mesh, such as the acoustic element transfer, the reciprocating compressor, the perforated plate, and the DOFs decoupling.

vibra/interface/numeric_checks/validators.py
```diff
@@ -0,0 +1,37 @@
...
+class StrictDoubleValidator(QDoubleValidator):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def validate(self, string: str, pos: int):
+
+        # if the string contains a comma, we reject
+        if "," in string:
+            return QDoubleValidator.State.Invalid, string, pos
```
andrefpf: We used to let the user type either `2.5` or `2,5`, since Brazilian users prefer the comma.

This could be kept by simply replacing it in the validator string, and keep the remaining checks, something like this:
```python
string = string.replace(",", ".")
...
```

vibra/interface/numeric_checks/validators.py
```diff
@@ -0,0 +1,37 @@
...
+                return QDoubleValidator.State.Invalid, string, pos
+
+        if is_numeric(string):
+            # check minimum value
+            if float(string) < self.bottom():
+                return QDoubleValidator.State.Invalid, string, pos
+
+            # check maximum value
+            if float(string) > self.top():
+                return QDoubleValidator.State.Invalid, string, pos
```
andrefpf: I am not sure if this always works as expected.
For example, if `self.bottom() == 3` and you try to write `100` the box will be blocked.

Even though `100 > 3`, the first digit is not `1 < 3`, therefore it will be marked as invalid.
The solution may be the `QDoubleValidator.State.Intermediate`.
It does not accepts the result, but allows user to continue typing.

vitorslongo:
When we try to import a new geometry with a previous analysis, this info is still kept: BC, mesh, fluid etc

https://github.com/user-attachments/assets/75c402a3-79d5-4aa2-a56c-47c0ae2205d5

vitorslongo (changes requested):

JacsonGilVargas (approved):
I've conducted several tests, and everything worked fine. The code is now better structured, without losing the performance achieved so far. Excellent work, @andrefpf. May God bless us.

PR #330, opened 2026-03-10
fixes symbols on surfaces with insufficient curvature

vini-ribeiro:
Surface normal calculations were slightly off on surfaces with minimal curvature. This PR use an 'if' to check these cases.


andrefpf (approved):
LGTM

PR #342, opened 2026-04-15
Fix cad-widgets conflict with vtk

andrefpf:


vini-ribeiro (approved):
LGTM! 🪨

vitorslongo (changes requested):
This did not work for me, the same error appears.

andrefpf:
It does not always works as expected =(

PR #356, opened 2026-05-06
Add loading window while hiding selection

JacsonGilVargas:
This pull request shows the loading window during the rendering process after hiding or showing selections. It also adjusts an acoustic model validation script file.


andrefpf (approved):
Everything works perfectly

PR #369, opened 2026-05-15
Unifies rendering toolbar with the camera toolbar

R0drigoSchwartz:
This PR contains the unification of the rendering toolbar with the camera toolbar.

Before:

image

After:
image

If users prefer, they can simply move the view toolbar and position it wherever they want.


JacsonGilVargas (approved):
I've run some tests, and everything worked fine. Excellent work, @R0drigoSchwartz. As a suggestion for future updates, we could replace the animation toolbar with a widget placed below the field plot. This approach will make the interface clean and more user-friendly.

PR #374, opened 2026-05-25
Avoid blank structures in the mesh workspace

### LINKED ISSUE #372: Avoid the blank structures in the mesh render
We should call the mesher setup interface whenever the mesh workspace is called after importing or drawing a geometry, and the mesh hasn't been generated yet. This approach might avoid the blank structures in the mesh render.

JacsonGilVargas:
This pull request implements a check to confirm whether a mesh has already been generated before switching to the mesh workspace. If the mesh has not yet been generated, the MesherSetup user interface will open, and the mesh workspace will be updated after the mesh has been processed. Otherwise, if no mesh is generated, the workspace will be returned to the model setup workspace. This PR closes #372.

### DIFF
--- a/vibra/engine/model.py
+++ b/vibra/engine/model.py
@@ -299,6 +299,14 @@ class Model:
         cond_B = len(self.solution_steps_mask) != int(sum(self.solution_steps_mask))
         return cond_A or cond_B
 
+    def is_the_mesh_setup_defined(self):
+        if isinstance(self.geometry_path, str | Path):
+            if self.is_there_a_geometry_imported():
+                if not isinstance(self.mesh_setup, MeshSetup):
+                    return True
+
+        return False
+
     def is_there_a_valid_mesh(self):
 
         if isinstance(self.geometry_path, str | Path):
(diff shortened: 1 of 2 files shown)


vibra/engine/model.py
```diff
@@ -299,6 +299,14 @@ def has_spectral_content_been_modified(self):
         cond_B = len(self.solution_steps_mask) != int(sum(self.solution_steps_mask))
         return cond_A or cond_B
 
+    def is_the_mesh_setup_defined(self):
+        if isinstance(self.geometry_path, str | Path):
+            if self.is_there_a_geometry_imported():
+                if not isinstance(self.mesh_setup, MeshSetup):
+                    return True
+
+        return False
```
andrefpf, line 308: This method is not being used?

andrefpf (approved):
Nice feature, it is much better now.

 I just did not understood the `is_the_mesh_setup_defined` method.

vibra/engine/model.py
andrefpf: I already saw it, sorry!

PR #392, opened 2026-05-27
show last 5 traceback lines in window

vini-ribeiro:
This PR changes which traceback lines are shown in the error window and how many are displayed when an exception is caught.


andrefpf (approved):
Good job. Can you also implement this for Open Pulse?

PR #414, opened 2026-05-29
Replace animation toolbar with widget

### LINKED ISSUE #398: Improve the layout of the toolbars

**Goal**: Improve the toolbars layout and UX.

**Tasks**:

- **Reduce toolbar rows to two**: currently there are 6+ toolbars (main, view, mesh, geometry, analysis, animation, workspace) arranged in a way that leaves empty/awkward frames when their contents are hidden. Reorganize so only two rows of toolbars are displayed.

- **Avoid empty toolbar frames**: today, when a toolbar isn't applicable (e.g., the animation toolbar outside of `AnimationRenderWidget`), only its contents are hidden, leaving an empty frame to preserve layout size. Find a cleaner approach that doesn't leave visible empty frames.

- **Move the animation toolbar into the analysis widget**: Users have been confusing the animation button with the "run analysis" button. Relocating the animation controls into the analysis widget (as shown in the topic below) should clarify the distinction and improve the workflow.

### What is the use case?

At the moment, the toolbars layout isn't very good.

We have a total of 6 toolbars (main, view, mesh, geometry, analysis, animation, and workspace), but they are displayed in a strange way.

To avoid displaying unnecessary toolbars, we have now hidden all their content, as shown in the image below:

Image

In this image, we are hiding the animation toolbar, which should only be displayed when using the `AnimationRenderWidget`.

Here is an example of the animation toolbar displayed, when the `AnimationRenderWidget` is active:

Image

It's good to have this approach of showing the toolbar only when it can be used. However, to avoid rendering with different sizes, we don't hide the entire toolbar, only its contents. This makes the interface a little awkward, because we stay with an empty frame:

Image

A good approach would be to keep only two rows in the toolbars. Additionally, it would be interesting to move the animation toolbar to the analysis widget, like in the image below:

Image

With this approach, we will improve the interface layout and user experience, since many users have reported confusing the animation button with the button to run the analysis.

Image

JacsonGilVargas:
This pull request replaces the current animation toolbar with an equivalent widget to be embedded on the field plot-related widgets.

<br>
Current animation toolbar:
image

<br>
<br>
Proposed animation widget:

image


R0drigoSchwartz (dismissed):
I checked the code and ran some analyses to observe the behavior of the animation widget, and everything seems to be working correctly. Now, the Vibra interface is much better and cleaner.

Great job @JacsonGilVargas! 👏

R0drigoSchwartz (approved):

andrefpf (approved):
As far as I tested, everything looks OK.

I just noticed that the Suction Silencer example is not running.
The same problem also happens in the main and dev branches, so I think this can be resolved in another PR.

PR #420, opened 2026-06-01
Reduce intialization time

### LINKED ISSUE #419: Slow startup

It is more common when using the installer, but after some branch changes it is also happening when running directly from the code. 
I measured about 17 seconds from the first click to the appearance of the splash screen in my work laptop. 
This is probably related to the import time of some libraries.

I think it is easy to calculate this initial time and raise a warning if it is too high. 
May be useful to catch the problematic libraries right when they are introduced.

### Mockups / Examples

_No response_

### Contact Details

_No response_

andrefpf:
This PR reorganizes a few imports to make the software initialization feel faster.
The time to start using the software is about the same, but the Splash Screen is appearing earlier.
We improved from about 16 seconds to only two.

I am not sure if it is a fluke, please let me know if the software is still slow.

Closes #419

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ dev = [
     "pytest-datadir>=1.7.2,<2",
     "invoke>=2.2.0,<3",
     "pyinstaller>=6.20.0",
+    "tuna>=0.5.15",
 ]
 
 [build-system]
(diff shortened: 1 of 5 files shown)


JacsonGilVargas (approved):
I loaded the application a few times, and the loading time is basically the same as the dev branch. However, the splash screen is displayed in less than 1 second after the application starts. Excellent work, @andrefpf.

R0drigoSchwartz (approved):
Nice job @andrefpf!

PR #430, opened 2026-06-02
Run analysis in a subprocess

### LINKED ISSUE #72: Suggestion: Stop or cancel button
As a user of other numerical simulation software, I find it interesting that Vibra/OpenPulse has a button to stop the analysis that is being performed by the user.

## For example

Imagine that you have configured a harmonic analysis and started it to run, but in the middle of the analysis you remember that you forgot to add some boundary condition, for example. Instead of waiting for the analysis to finish, you cancel it, adjust the model configuration and start it to run correctly.

vini-ribeiro:
This PR adds support for running analyses in a subprocess. The implementation uses `subprocess.Popen` to start a child process that runs a script responsible for reading the project from the `working_directory` and executing the analysis.

Points to be evaluated:

- Whether all analyses continue to run correctly.
- Whether exceptions raised inside the subprocess are properly propagated to the parent process.
- Whether these exceptions are displayed correctly in the terminal and in the interface error window.
- Whether the analysis execution time remains stable and within an acceptable range.


andrefpf:
I think it is valid to inform that @vini-ribeiro used `subprocess.Popen` instead of `multiprocessing.Process` because it appears to offer a considerable performance advantage.

vibra/utils/subprocess/solver_subprocess_error.py
andrefpf, line 1: We have a file in `vibra/errors.py` with many small classes like this one. You can move it there.

vibra/utils/subprocess/subprocess_handler.py
```diff
@@ -0,0 +1,108 @@
...
+class SubProcessStatus(Enum):
+    SUCCESS = auto()
+    INTERRUPTED = auto()
+
+
+class SubProcessHandler:
+    """Run the configured project analysis in a separate Python process."""
+
+    def __init__(self):
+        self.process_script = VIBRA_DIR / "utils/subprocess/analysis_subprocess.py"
```
andrefpf: You could allow the class to receive a generic file path. It will probably be used in other occasions as well. 
```python
def __init__(self, path: Path):
    ...
```

vibra/utils/subprocess/subprocess_handler.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def run_analysis_in_subprocess(self) -> SubProcessStatus:
+        self._subprocess = None
+        self._interrupted = False
```
andrefpf: Type `self._subprocess` to make the linter happy.

```suggestion
        self._subprocess: subprocess.Popen | None = None
        self._interrupted: bool = False
```

vibra/utils/subprocess/subprocess_handler.py
```diff
@@ -0,0 +1,108 @@
...
+            stderr = self._subprocess.stderr.read()
+            logging.error(f"Solver subprocess exited with code {self._subprocess.returncode}")
+            raise SolverSubprocessError(
+                returncode=self._subprocess.returncode,
+                stderr=stderr,
+            )
+
+        return SubProcessStatus.SUCCESS
+
+    def _read_pipe_lines(self, pipe, line_queue: Queue[str]):
```
andrefpf: Use  `from typing import IO` to hint the type of `pipe`.

andrefpf:
The stop button is not appearing in the structural modal analysis from the examples, as shown in the image bellow:

image

andrefpf:
For the Harmonic Acoustic analysis of the suction silencer, with element size of 20 mm, the time dropped from ~717 seconds to ~323 seconds.

Very good.

andrefpf (dismissed):
Great work. As far as I can tell everything is working perfectly.

JacsonGilVargas (approved):
I've conducted a couple of tests and everything worked fine. Now, we can stop the heaviest processing in all analyses. This is very useful, especially when solving the eigenproblems. In terms of performance, I noticed a time saving of around 10% with similar RAM usage, which is awesome. Excellent job, @vini-ribeiro.

Issue #402, opened 2026-05-27
Copy porous material for other volumes


andrefpf:
### What feature would you like?

When the porous material is already defined for one body there is no simple way to applying the same material to another volume.

<img width="2880" height="1620" alt="Image" src="https://github.com/user-attachments/assets/2a177d0c-9512-4a5c-98a0-713b27efd13f" />

JacsonGilVargas:
The functionality to copy the input parameters of the porous material model has been implemented for quite some time. The porous material parameters are copied whenever a volume to which a porous material has already been assigned is selected. Moreover, this functionality is present in almost all other model configuration interfaces, @andrefpf.

PR #432, opened 2026-06-03
Enhancements to the porous material UI

JacsonGilVargas:
This pull request modifies the porous materials user interface to make configuring porous material models easier. It is now possible to adopt the standard constants from the Delany-Bazley and Delany-Bazley-Miki models, as well as a user-defined set of constants. Although this functionality has been implemented for quite some time, I would like to emphasize the possibility of loading the input data for porous materials by clicking on the volume where a porous material has already been defined. It also resolves the issue #402.


andrefpf:
In the first line were it reads "Effectives complex wave...", the spelling is correct or it should be "Effective"?

image

andrefpf:
In my screen the default window size ommits part of the content.
Although there is the scroll bar, looks like there is nothing below.

image


Maybe it should start like this?

image

andrefpf (dismissed):
Everything seem to be working fine.
I added a few comments in the PR, but feel free to ignore them if they are not applicable.

JacsonGilVargas:
Actually, there was a spelling error related to the word "Effectives", @andrefpf. As for the initial window size, I just added a button to control when the advanced controls for the porous material plots will be displayed. These plots are designed for advanced users, so I'll let them be disabled by default.

andrefpf (approved):
Nice!

PR #435, opened 2026-06-04
Minor adjustments for delete properties feature

JacsonGilVargas:
This pull request proposes to standardize and enhance the delete feature you developed, @vini-ribeiro. The main idea is to make the user experience more dynamic and avoid stacking multiple pop-ups on screen. While testing the PR you opened, I wondered if the strategy of obtaining user confirmation every time a property is removed is the most appropriate. I'm not sure, but it doesn't seem to be the best approach from a user experience perspective. What do you think about it? I took the opportunity to fix minor problems I found while reviewing the code. Please, @vini-ribeiro, let me know if you agree with the proposed updates.


vini-ribeiro (approved):
Great! I agree with all the proposed changes.

Issue #383, opened 2026-05-26
Analysis type is being unexpectedly modified


andrefpf:
### What happened?

If we configure a harmonic analysis

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/787f21fa-ba6d-4230-ae3d-99e74433aa9d" />

Then change it to modal.

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/c41ed5e3-621b-4c6a-bc6a-603c1f05d5e3" />

And run the analysis without setting the modal configuration.

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/995a3f77-2a48-488d-91db-e6dbd94ba67a" />

The analysis type falls back to harmonic and runs without any errors. I think it is a bit misleading. 
The software should instead show an error and require the user to configure the modal analysis.
Open pulse does exactly it.

### What operational system are you seeing the problem on?

Windows

### Version

dev branch

### Relevant log output

```shell

```

Issue #423, opened 2026-06-01
Make the analysis setup button stand out more


vitorslongo:
### What feature would you like?

It was asked in the UP course that we made this button more proeminent. We could just replicate what we did with the workspaces buttons

<img width="497" height="145" alt="Image" src="https://github.com/user-attachments/assets/01ddf20b-fefa-4631-9013-48d0eb168b25" />

PR #436, opened 2026-06-08
Add improvements to the toolbars

JacsonGilVargas:
This pull request uniformizes the toolbars' appearance and adds some improvements to the analysis setup button after changing the analysis type or physical domain. Now, if you have an analysis configured and change the analysis type or physical domain, the Run Analysis button will stay deactivated until a valid analysis setup is configured. It resolves issues #383 and #423.


vitorslongo (dismissed):
Nice work! The interface looks much better now. I've ran some tests and everything seems to be working perfectly.

andrefpf (approved):
LGTM

Issue #379, opened 2026-05-26
Verify anechoic termination during TL


andrefpf:
### What feature would you like?

When we try to plot a Transmission Loss between two faces that to not have radiation impedance applied the following message appears.
In Open Pulse a better message appears, we should do the same here.
 
<img width="2880" height="1620" alt="Image" src="https://github.com/user-attachments/assets/bd90328b-8164-4c4c-ab09-60573fe771b9" />

### What is the use case?

The error is pretty common, but the message is not clear enough.

### Mockups / Examples

[Captura de tela de 2026-05-26 15-19-03.png](https://github.com/user-attachments/assets/ad76e714-8067-4849-bf47-83abcac86299)

PR #438, opened 2026-06-09
Enhance the verification requirements for transmission loss calculation

JacsonGilVargas:
This pull request enhances the verification requirements for transmission loss calculation. A help button has also been added to display a message regarding the requirements for transmission loss and noise reduction calculations. Even though it is recommended to have a minimal background in the Acoustics field, the proposed updates should prevent the most common user errors when calculating transmission loss.

This PR closes issue #379.


andrefpf (approved):
LGTM

PR #439, opened 2026-06-09
Mesh subprocess

vini-ribeiro:
This PR adds an option to generate meshes using a subprocess. The implementation uses Python's ```subprocess.Popen``` to run the mesh generation in a child process.
It also enables Gmsh multithreading, allowing Gmsh to use all available threads during mesh generation.
This PR also changes the behavior of the exception widget. The widget now has priority over other windows and is displayed on top.


vibra/engine/mesher/mesh.py
```diff
@@ -209,9 +209,14 @@ def new_load_cad(self, path: str | Path, mesh_setup: MeshSetup):
...
             gmsh.option.set_number("General.Verbosity", 0)
-            # gmsh.option.set_number("General.NumThreads", threads)
             gmsh.option.set_number("Geometry.Tolerance", mesh_setup.geometry_tolerance)
 
+            n_thread = 0
+            gmsh.option.set_number("General.NumThreads", n_thread)
+            gmsh.option.set_number("Mesh.MaxNumThreads1D", n_thread)
+            gmsh.option.set_number("Mesh.MaxNumThreads2D", n_thread)
+            gmsh.option.set_number("Mesh.MaxNumThreads3D", n_thread)
+
```
andrefpf, line 202: We tested this in another moment and it segfaulted sometimes.
Maybe the gmsh guys already fixed it, so lets give it a try.

andrefpf (dismissed):
As far as I tested it seems to be working as expected and the code looks fine

JacsonGilVargas (approved):
As far as I could test, everything seems to be working fine. I've conducted some tests to generate meshes on problematic geometries, and the code didn't crash. Excellent job, @vini-ribeiro.

PR #442, opened 2026-06-09
Fix symbols position in mesh refinement

### LINKED ISSUE #140: Position of the symbols is influenced by the mesh refinement
As the image below shows, the symbols on the left are not centered.
Image

From the mesh, it's clear that the refinement occurs on the side where the symbols are not properly aligned.
Image

GuilhermeSPierri:
This pull request fixes #140.


vini-ribeiro (approved):
Great work! 🚀

vitorslongo (approved):
Smart move! Seems to be working

Issue #422, opened 2026-06-01
Show element edges in bad elements plot


vitorslongo:
### What feature would you like?

Sometimes it is clearer for the user, when debugging the mesh, to enable the edges view in the bad elements plot.
As of now, the plot is by default like this:

<img width="1263" height="755" alt="Image" src="https://github.com/user-attachments/assets/b76b7ee9-29ca-4780-a019-ec2378e28e13" />

And enabling the edges makes all of them appear, making it hard to actually understand something:

<img width="1263" height="755" alt="Image" src="https://github.com/user-attachments/assets/a03d57f8-0cae-4624-8217-818193682348" />

With the edges, it's actually clear why the element is ill-conditioned, but we can't see what's happening on the elements further inside the mesh:

<img width="1263" height="755" alt="Image" src="https://github.com/user-attachments/assets/d0e79a1b-33cf-43e9-b7eb-e17bb75bc69f" />

PR #443, opened 2026-06-10
Use different visualization filter

### LINKED ISSUE #381: Use a different VisualizationFilter for each workspace

Visualization filter is a dataclass that holds information that is controlled by this part of the visualization toolbar:

Image

There is a single visualization filter in the main window, but it causes some inconveniences.
Points/Nodes are commonly activated for the geometry workspace, but deactivated on the mesh and results workspaces.
Lines/Edges are commonly activated for the geometry and mesh workspaces, but deactivated in the results workspace.

Currently we set a coherent default value during workspace changes, but they may conflict with the user intent.
Therefore, the user must have control of which entities should be visualized on each workspace, and these choices must be remembered across workspace changes.

Image

Image

### Mockups / Examples

_No response_

### Contact Details

_No response_

andrefpf:
- Closes #381 
- Recovers already merged colors from splash screen
- Implements a few fancy classes that may (or may not) be used later for different result plots


andrefpf:
@vitorslongo I am marking you to review this because this PR can possibly mess with the "bad elements" and "disconnected nodes" plots. Since you implemented this I think you can test it properly.

vitorslongo (dismissed):
I've noticed some behaviours that could use some attention.
1. The "plot bad elements" button does not automatically hides all mesh edges (when lines view is enabled), thus making it hard to actually see anything:

image

However, maybe we don't need to fix it now since this will be dealt with on the PR that fixes https://github.com/MOPT-UFSC/VIBRA/issues/422. I'll give this some attention soon, but just to let you know that something along these lines will change.

2. Also, whenever we click on the mesh setup, or close the mesh setup window while plotting the bad elements, it goes back to the model setup workspace. IMO this should only happen when we click on the local refining tab. For these other cases this behaviour is annoying.

Specifically the bad elements plot is ok, no bugs or anything that was broken.

Since I don't know if going back to the setup workspace was caused by the changes you've made here, I'll approve it and if actually not we can create an issue and change this separately. The same goes to fixing https://github.com/MOPT-UFSC/VIBRA/issues/444, I don't know if this is related to the changes of this PR.

vitorslongo:
For the disconnected nodes, the only geometry i had that raised that error seems to be working now (mesh and results look ok), but looking at the code, what you've done probably does not affect its operation.

andrefpf:
> Also, whenever we click on the mesh setup, or close the mesh setup window while plotting the bad elements, it goes back to the model setup workspace. IMO this should only happen when we click on the local refining tab. For these other cases this behaviour is annoying.

The first behavior was resolved. 

The second one is not specific from this PR. 
You can create an issue for it and we discuss the solution in another moment.

vitorslongo (approved):

PR #448, opened 2026-06-11
Standardize model properties UI main buttons

JacsonGilVargas:
This pull request resolves issues #386, #453, and #454.


andrefpf (changes requested):
When I try to add a DOF decoupling the "ok" button is not working.

Image

Also in the DOF decoupling, when we use "apply" it in fact applies, but an error appears.
It is probably an unintended callback regarding the "Selected ID" combobox.

image

andrefpf:
When I open the "VIscous-thermal Loss Model" item in the "acoustic_metamaterial_model" example, the following error message is raised:

image

andrefpf:
If I try to add an invalid condition pressing "OK", like an empty field in "Acoustic Pressure" for example, an error message appears (as expected).

The problem is that it also closes the input window when the message is disregarded.
The input window should only be closed if the attribution succeeds. 

This problem happens in most input windows.

image

andrefpf:
I also noticed the bugs reported in #454 and #453. 

Since they also occur in the `dev` branch I preferred to report them as a separated issues, even though you may chose to fix them here as well.

vitorslongo (dismissed):
I've encountered no issues, the UI is much much better now 👏

andrefpf (dismissed):
This time I could not find any problems with the interface.
Everything feels much more consistent with these changes.

Good job!

andrefpf (approved):

Issue #424, opened 2026-06-01
Show the face on which we apply anechoic termination


vitorslongo:
### What feature would you like?

When adding an anechoic termination to an entity, only the volume ID is shown. It was stated in the UP course that this is confusing.

<img width="1287" height="650" alt="Image" src="https://github.com/user-attachments/assets/38db5735-8326-4f6c-82d0-97e4fbcf9296" />

Is there a reason for this to be this way?

PR #459, opened 2026-06-12
Refactor anechoic termination user interface

JacsonGilVargas:
This pull request modifies the anechoic termination user interface files to guide users when configuring this model property. A couple of checks and filters have been added to prevent assigning this property to an internal surface. Since anechoic termination does not require numerical inputs, a table with specific impedance values ​​for the selected surfaces has been added to assist users. (Closes issue #424)
<br>
image


vitorslongo (dismissed):
Nice, this'll make it easier for not so experienced users 👏

andrefpf (approved):

PR #463, opened 2026-06-15
Add timer utils

### LINKED ISSUE #394: Add time warnings

It is very hard to keep the software consistently fast.
Most optimizations are based on some constraints that with time may be modified and are lost.
We should have a few time utilities to ensure that optimizations are not lost over time.

Currently my suggestions are the following:
- Store a timestamp in the software startup and add warnings if the delta time from startup to a few keypoints exceeds a maximum time limit
- Create a decorator to use in critical functions that should execute instantly. This decorator will log a warning whenever the function takes more than a maximum time limit

### Mockups / Examples

_No response_

### Contact Details

_No response_

andrefpf:
Add a few utilitary functions to help with execution time optimization tasks.
A few functions and decorators were added to sections of code optimized recently.

closes #394


R0drigoSchwartz (approved):
These timer functions are great and will be very useful. Good job, @andrefpf

PR #467, opened 2026-06-17
Enhance how to configure element type and add warning icons to the results items

### LINKED ISSUE #465: Add warnings whenever a solution is available and the model configuration is modified

We might add warnings, via icons and tooltips, to make the user aware whenever a solution is available and the model configuration is modified. An exception to this rule is the model configuration that results in DOF decoupling (DOF decoupling, transfer impedance, and perforated plate). For these model inputs, we must reset the solution to prevent a mismatch between the model solution DOF and that of the just-updated mesh.

### Mockups / Examples

_No response_

### Contact Details

_No response_

### LINKED ISSUE #468: Error when solving the structural_discharge_system.vibra project example
### What happened?

An error occurred when trying to solve the "structural_discharge_system.vibra" project example. The error is related to a misinterpretation of the structural element type. This example uses a customized MeshSetup that is causing problems when mapping the structural element type.

Windows

### Version

dev branch

### Relevant log output

```shell

```

### Upload screenshots

_No response_

### Contact Details

_No response_

JacsonGilVargas:
This request aims to improve how the element type is configured in the model and to adjust the terminology for attributes and methods. The proposed modifications resolve an error detected when solving the structural_discharge_system example. Additionally, I took the opportunity to implement a feature that adds warning icons to the results menu items whenever the model solution is out of date with respect to the configuration.

Note: I know this isn't the recommended practice @andrefpf, so I apologize in advance. I continued working on the same branch and, before I knew it, I had already sent several updates and didn't want to lose the progress I had made.


vitorslongo (dismissed):

vibra/engine/elements/element_type.py
andrefpf, line 1: I can not comprehend why are these types so redundant. We could only have ElementTopology and all required information would be correctly encoded.

validation_files/validate_acoustic_element_hex20_harmonic.py
```diff
@@ -118,10 +113,12 @@ def load_external_mesh_and_solve():
         molar_mass=molar_mass,
     )
 
-    ## assign the created fluid
+    ## intialize the model
     model = Model()
+    model.set_element_type(element_geometry="hexahedral", element_order="quadratic")
```
andrefpf: Please correct if I am wrong. But the topology isn't determined by the mesh itself?

For example, if I configure in the `Model` a topology that isn't compatible with the mesh, the software will probably crash. Therefore, this attribute should be inside the `Mesh` class.

When the Mesh is generated by gmsh it is only a matter of copying the attribute from the MeshSetup.
But otherwise, if the mesh was already generated externally, we only need to infer its topology from the element connectivity.

vibra/interface/menus/common_menu_items.py
```diff
@@ -151,14 +151,17 @@ def set_property_name(self, name: str):
...
-            font = QFont()
-            font.setBold(True)
-            self.setFont(0, font)
-            self.setForeground(0, Color(*color_names.YELLOW.to_rgb()).to_qt())
+            if update_item_color:
+                font = QFont()
+                font.setBold(True)
+                self.setFont(0, font)
+                self.setForeground(0, Color(*color_names.YELLOW.to_rgb()).to_qt())
+
```
andrefpf, line 175: I am not so sure about this part.

Don't you think it too inconsistent to have different warning styles for each menu?
I don't really have any preference, but we should stick with only one style.

Image

Image

vibra/interface/menus/results_viewer_items.py
```diff
@@ -246,3 +249,41 @@ def set_theme(self, theme : str):
...
+            top_level_item.setFont(0, font)
+            top_level_item.setForeground(0, Color(*color_names.RED_5.to_rgb()).to_qt())
+            warning_icon = QIcon(str(ICON_DIR / "model_setup_items/warning_yellow.png"))
+            top_level_item.setIcon(0, warning_icon)
+
+        else:
+            # Resets data to default
+            top_level_item.setData(0, Qt.FontRole, None)  # reset color
+            top_level_item.setData(0, Qt.ForegroundRole, None)  # reset color
+            top_level_item.setData(0, Qt.DecorationRole, None)
```
andrefpf: This method can be moved to `vibra.interface.menus.common_menu_items::TopTreeWidgetItem`, since there is already an equivalent implementation in `ChildTreeWidgetItem`.

This way both items can be highlighted with `item.set_warning(value)`

andrefpf (dismissed):
I left a few comments that could simplify the implementation.

But since it seems to be working I will approve anyway and let you decide if it should be modified in this PR or not.

vibra/interface/menus/results_viewer_items.py
JacsonGilVargas: Done!

vibra/interface/menus/common_menu_items.py
JacsonGilVargas: I don't see any problem with these differences, since the highlighted texts in the model setup denotes an obligation, and in the model results we're just advising the user to be careful when analysing results. Anyway, I'll show the proposal to other users to get more opinions or even suggestions to improve this feature.

validation_files/validate_acoustic_element_hex20_harmonic.py
JacsonGilVargas: You're right, @andrefpf, this setup might be simplified.

andrefpf (dismissed):

andrefpf (approved):

PR #474, opened 2026-06-18
Add deletion shortcuts

### LINKED ISSUE #455: Improve removal of single selected properties

When we select an entity containing a single property and press delete, the whole "selection" window appears asking the user to select which property they want to remove.
Since there is not much to select, it is a bit annoying and should not be required.

Image

Despite that, the confirmation to remove the property should still be presented.

Image

### Mockups / Examples

_No response_

### Contact Details

_No response_

### LINKED ISSUE #456: Select all properties in the removal window

When the window is opened none of the items is selected. 
I think selecting all items is a better default behavior, since the user may want to remove all selected properties.

The whole selection could be removed with 3 keystrokes: del, enter, enter ❤️ 

Image

### Mockups / Examples

_No response_

### Contact Details

_No response_

vini-ribeiro:
This PR updates the delete window behavior:

- When only one property is selected, the confirmation is shown directly without displaying the options table.
- All properties are selected by default.

Closes #455 
Closes #456

### DIFF
--- a/vibra/interface/model_inputs/general/choose_property_to_delete.py
+++ b/vibra/interface/model_inputs/general/choose_property_to_delete.py
@@ -21,6 +21,10 @@ class ChoosePropertyToDelete(ChoosePropertyToDelete_UI):
         if not self.properties_formated:
             return
 
+        if len(self.properties_formated) == 1:
+            self.tableWidget.selectAll()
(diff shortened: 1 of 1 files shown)


andrefpf (approved):
GREAT!

PR #478, opened 2026-06-19
Remove unnecessary code

andrefpf:
Deletes a bunch of files and variables that were not being used.
I started this to remove the two unused methods from mesh.py as we previously talked this week, but I took the chance to remove a bunch of deprecated code.


JacsonGilVargas (approved):
I've performed a couple of tests, and everything worked well. Excellent code cleaning, @andrefpf.

Issue #475, opened 2026-06-19
Add graphs for vibration velocity and acceleration fields


JacsonGilVargas:
### What feature would you like?

It would be interesting to have the field plots for vibration velocity and acceleration. We might keep the same results widget and add a new control to select the result type: displacement, velocity, or acceleration.

PR #481, opened 2026-06-19
Generalizes the structural response field plots #475

JacsonGilVargas:
This pull request generalizes the current structural displacement field graph, also allowing the structural results to be plotted in terms of vibration velocity and acceleration. Some methods have been reformulated to generalize structural graphs to multiple data types.

<br>
image


andrefpf (changes requested):
I found the following division by zero bug when trying to plot the velocity and acceleration result for the 0 Hz frequency of an example cylinder.

The problem does not seem to be specific from the derivated plot, it was only highlighted by the derivation.

Regarding the code, everything seems smartly done. Good job.

Image<br>

> I've just resolved the division-by-zero error, @andrefpf. Please repeat your tests and let me know if everything is working fine.

andrefpf (dismissed):

vitorslongo (approved):
Everything seems to be working perfectly. Nice addition!

PR #491, opened 2026-07-04
Update the element topology based on face connectivity as well

JacsonGilVargas:
This pull request corrects an error found when updating the element topology based on connectivities. The element type has not been updated in the absence of solid element connectivities. Now, the element topology will be updated based on face connectivities if there are no solid elements.


vitorslongo (approved):
Seems to be working perfectly

andrefpf (approved):
Working fine, good job

PR #493, opened 2026-07-20
Fix icons

### LINKED ISSUE #384: Icons are not always being painted
### What happened?

Sometimes the icons are not being painted. We should probably stop painting the icons and dynamically load them with the correct colors. 

Image

Windows

### Version

dev branch

### Relevant log output

```shell

```

### Upload screenshots

_No response_

### Contact Details

_No response_

R0drigoSchwartz:
This PR contains an alternative approach to coloring the icons by switching between different resource files that already contain the colored icons.

Now, we can paint the icons by running a script named `paint_icons.py`, which is located in the `scripts` folder. This script also executes the `invoke ui-compile` command, which generates the new resource files containing the new icons.

Closes #384


vibra/interface/data/icons/dark_theme/geometry_icon.png
andrefpf, line 1: is it correct?

scripts/paint_icons.py
```diff
@@ -0,0 +1,54 @@
...
+def save_icon(icon_path: Path, icon_image: Image) -> None:
+    icon_path.parent.mkdir(parents=True, exist_ok=True)
+    icon_image.save(icon_path)
+
+
+def paint_icons():
+    for dir, color in zip(ICONS_DIRS, COLORS):
+        icons_to_paint = get_icons_to_paint(dir)
+
+        for icon in icons_to_paint:
```
andrefpf:
```suggestion
        for icon in get_icons_to_paint(dir):
```

scripts/paint_icons.py
andrefpf, line 1: nice code

scripts/paint_icons.py
```diff
@@ -0,0 +1,54 @@
...
+            img = Image.open(icon).convert("RGBA")
+            img_data = np.array(img)
+
+            mask = img_data[:, :, 3] != 0
+            img_data[mask] = color.to_rgba()
+
+            painted_icon = Image.fromarray(img_data)
```
andrefpf, line 41: A nice addition would be to resize all icons to be of the same size, like 256x256 or something like this.

We need to be careful to avoid stretching the non-rectangular images.
Probably do it in another PR anyway.

andrefpf (dismissed):
Everything seems to be working now, and this approach looks much better than the previous one.

vibra/interface/data/icons/dark_theme/geometry_icon.png
R0drigoSchwartz: No, I will fix it!

scripts/paint_icons.py
R0drigoSchwartz: Ok! Thanks by the suggestion.

andrefpf (dismissed):

JacsonGilVargas (approved):

PR #494, opened 2026-07-20
Acoustic pressure waveform field

andrefpf:
Implements the plot for transient pressure. Now the animation looks reasonable.
I also edited the code of existing plots, so it is worth testing them as well.

I will mark some attention points directly in the code.


vibra/engine/postprocessing/acoustic_postprocessing.py
```diff
@@ -139,20 +136,72 @@ def compute_acoustic_pressure_field(
...
+                max_value = np.max(tmp)
+
+            case _:
+                min_value = np.min(waveform)
+                max_value = np.max(waveform)
+
+        return acoustic_pressures, min_value, max_value
+
+    @cache
+    def compute_multiple_ifft(self):
```
andrefpf, line 174: @JacsonGilVargas I moved this method here to match the behavior of the other analysis. Note that the `cache` decorator avoids the recomputation, just like it did before.

vibra/engine/postprocessing/acoustic_postprocessing.py
```diff
@@ -139,20 +136,72 @@ def compute_acoustic_pressure_field(
...
+                acoustic_pressures = np.imag(acoustic_pressures)
+                tmp = np.imag(waveform)
+                min_value = np.min(tmp)
+                max_value = np.max(tmp)
+
+            case _:
+                min_value = np.min(waveform)
+                max_value = np.max(waveform)
+
+        return acoustic_pressures, min_value, max_value
```
andrefpf, line 171: I tried to implement this based on the other method, but it is probably wrong.
@JacsonGilVargas please verify.

JacsonGilVargas (approved):
Well done, @andrefpf.

PR #497, opened 2026-07-21
Disable points in geometry render widget

### LINKED ISSUE #495: Change the default visualization filter presets
The current view filter presets are displaying the geometry points,

Image

<br> but they should initialize without showing them.

Image

The strategy of not showing the point is aligned with those adopted in various commercial software.

andrefpf:
Closes #495


JacsonGilVargas (approved):
Excellent job, mate.

PR #499, opened 2026-07-22
Disable lazy solution

andrefpf:
This feature was doing much more harm than good, because it wasn't very suited for our use case.
As far as I am concerned, lazy algorithms are useful mainly in two cases:
- if the content being loaded is only necessary in a few cases;
- when loading everything at once has a high execution time penality, but the parts are relativelly fast. In this case spreading the calls makes the application feel more fluid.

None of this makes sense for our application. 
Regarding the first point, the user will likely want to see many frequencies/modes in sequence, or combine them in a transient plot (comming soon =).

For the second point, we were "lazily loading" so much data at once that the time penality was not being spreaded, it was only beind moved to the most random places, and making the bottlenecks hard to debug.

On the other hand, the lazy writer has its advantages, in cases of interruptions for example.
The hardest part of this PR was decoupling the "lazy loader" from the "lazy writter", since the latter is way more useful than the former.


andrefpf:
@JacsonGilVargas I took a bit of an adventure in the HarmonicSolver, so please check if the code is still making sense and if I did not broke any analysis.

Note that the "LazyWritter" was kept, so we can resume the analysis if it was stopped in the way.
But while it writtes in the file, it also writes into an in-memory buffer, so we don't lose performance.

JacsonGilVargas (commented):
I've performed some tests and noticed a huge bottleneck when loading project data of medium-sized models, @andrefpf. Consider medium-sized models to be those with more than 100k degrees of freedom (DOFs) and more than 200 solution steps. After opening a medium-sized model, the progress bar stuck at 85% for minutes.

Image

JacsonGilVargas (approved):
I've conducted some tests, and the time to load solution data has consistently decreased. The loading bottleneck seems to have disappeared, and no other issues have been found. Excellent enhancement, @andrefpf.

PR #509, opened 2026-07-27
Add estimated progress to eigenproblem

vini-ribeiro:
This PR estimates the number of times the `_matvec` method from `PardisoLinearOperator` will be executed, in order to provide progress feedback to the user.

It uses the formula `min(A.shape[0], max(2 * (n + 1), 20))`.

Please verify whether this formula is correct. I tested it and it seems to be working fine.


JacsonGilVargas (dismissed):
I've performed several tests, and the eigensolver progress bar worked very well, @vini-ribeiro. I've tested replacing 2 * n_modes + 1 with 3 * n_modes + 1, and it seems to approximate the actual total number of iterations better. Please, @vini-ribeiro, test the proposed multiplier and let me know if you judge it worth modifying. Before I forget, excellent work, mate.

JacsonGilVargas (approved):

PR #512, opened 2026-07-29
Improve efficiency of incident plane wave inputs

### LINKED ISSUE #511: Incident plane wave is very slow
### What happened?

The selection calback inside the Incident Plane Wave widget is painstakingly slow for medium sized project (60k+ nodes).
In an example used for debugging (the image below) it took up to 140 seconds (2 minutes and 20 seconds) for the largest surfaces, and around 20 seconds for the remaining surfaces.

Image

Windows

### Version

dev branch

### Relevant log output

```shell

```

### Upload screenshots

_No response_

### Contact Details

_No response_

andrefpf:
Fixes #511

In the same case that was tooking 140s now takes 7ms. 
About 20.000x faster.

To achieve such impovement a few things needed to be changed.
I created a new version of `mesh.get_element_face_normal` named `mesh.get_element_face_normal_batched`.
The idea is to send a 2D array at once instead of multiple 1D arrays.

I kept both methods because for this new version the expected connectivity includes the indexes, and I was not sure how to update the associated code that was calling the previous method.

The indexes were needed to retrieve the associated solid using `self.face_to_solid_element[i]`.
Inferring this information is very slow, and the code to populate these dicts is very optimized, so I preffered to use the existing data.

This PR also updates a few logging related stuff that helpped with debugging.


JacsonGilVargas (approved):
I've conducted several tests, and everything worked flawlessly. Furthermore, I observed a significant performance gain when calculating the normals of the 2D elements. Excellent work, @andrefpf.

PR #517, opened 2026-07-31
Feat transparency

R0drigoSchwartz:
This PR adds the transparency feature, which allows the user to view the internal parts of the analyzed geometry.

Here, I followed the same approach that we use on Open Pulse.


andrefpf (approved):
great

PR #523, opened 2026-08-04
Fix subprocess executable

andrefpf:
I modified a bit the SubprocessHandler to make it work for executables.
To create an executable run `uv run pyinstaller vibra.spec --noconfirm`, and test the executable in `./dist/Vibra/vibra.exe`.
I did not tested it on Linux, so please let me know if anything weird happens.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1716,7 +1716,7 @@ wheels = [
 
 [[package]]
 name = "vibra"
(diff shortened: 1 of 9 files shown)


vini-ribeiro (approved):
Great Work! Good insight in using flags to run vibra.exe as CLI 🚀

PR #526, opened 2026-08-04
Fix volume selection

andrefpf:


JacsonGilVargas (approved):
The volumes selection is now working flawlessly. Thank you very much, @andrefpf.

PR #531, opened 2026-08-05
Add mesh coarsening

### LINKED ISSUE #529: Add mesh coarsening

Coarsening complex geometries currently lacks targeted automation. Users are forced to manually select all entities they wish to exclude from coarsening to achieve localized refinement elsewhere—a process that is error-prone, inefficient, and creates significant friction on large-scale models.

### Mockups / Examples

_No response_

### Contact Details

_No response_

### LINKED ISSUE #530: Invalid mesh setup window appears behind mesh setup window
### What happened?

Sometimes, also, an error is shown that does not explain clearly why the setup is incorrect.

Windows

### Version

dev branch

### Relevant log output

```shell

```

### Upload screenshots

[Captura de tela 2026-08-05 134042.png](https://github.com/user-attachments/assets/cded5f3e-15c1-46b9-82d5-93be8954d1e0)

### Contact Details

_No response_

vitorslongo:
This PR implements an inverted mesh refinement, a long ago requested functionality that resurfaced recently. 
Also:
- Adds guardrails for raising the minimum mesh size beyond the maximum
- Adds mesh pytests related to mesh refining and coarsening
- Adds a more precise control for selecting entities to refine/coarsen. Now the user can input IDs in addition to selecting in the render


JacsonGilVargas (changes requested):
I've conducted some tests with the local mesh size controls and found two problems. The first is simple: whenever we configure an element size greater than the global mesh size a message is displayed informing the user about the local coarsening. We can remove this message, as it no longer makes sense after recent modifications. The second—and serious—problem concerns the entities typed by the user in the local control size. For instance, I typed a local size for a non-existing entity ID, and the code crashes when trying to generate the mesh.

<br>Image

<br>As a general rule, whenever a feature will work with typed inputs, we must check whether the typed entities exist before using them. Fortunately, this problem is easy to solve: simply call the "check_selected_ids" method of the Mesh class.

JacsonGilVargas (dismissed):
I ran a couple of tests, and the element size control seems to work properly now. Amazing contribution, @vitorslongo.

andrefpf (changes requested):
I do not think the functionality is working as expected =/

This is the base mesh with min/max elements of size 0.01 and 2.00:
Image

When I a coarsened volume 1, the configuration seems to have been inherited by the lines of the remaining volumes:
Image

When I added surface another coarsening to surface 9 it forgot about the coarsening of volume 1:

Image

vitorslongo:
Ok, I'm sorry. I've refactored all the coarsening logic and now everything seems to work, including simultaneously:
image

I've seen no more leaks to lines or points. Now every local size control follows the same logic, a much more robust approach.

JacsonGilVargas (approved):
Everything seems consistent now. Incredible feature, @vitorslongo.

andrefpf (approved):
I runned a few tests and everything seems to be working.
The code is also pretty good. Great job.

PR #533, opened 2026-08-06
Add normal symbols

andrefpf:


JacsonGilVargas (approved):
I ran some tests, and everything worked perfectly. Thank you very much for the support, @andrefpf.

PR #535, opened 2026-08-07
Clears the stylesheet of the line edits correctly

R0drigoSchwartz:
## What does this PR do?

This PR fix a bug that started to happen with the line edits after an update of the PySide6 dependency. To fix it, I created a function called `clear_style_sheet`, in the `vibra/interface/general/utils.py` file, that clears and update the style sheet of a widget correctly.


JacsonGilVargas (approved):
As far as I was able to test, everything worked perfectly. Excellent corrections, @R0drigoSchwartz.

PR #556, opened 2026-08-21
remove try/except in cli.py

vini-ribeiro:
This PR removes the try/except block in ```cli.py```, which was suppressing exceptions thrown in analysis/mesh.


andrefpf (approved):
fair enough

PR #557, opened 2026-08-22
Add volume suppression

### LINKED ISSUE #400: Suppress geometry volumes for analysis

In cases where we have big geometries but only want to simulate a certain part of it, it would be useful to suppress some volumes, making the analysis faster.

### Mockups / Examples

_No response_

### Contact Details

_No response_

vitorslongo:
This PR implements volume suppression. 
Whenever a volume is suppressed, it is not meshed and its BCs are not computed in the analysis. This can be particularly useful when dealing with large projects/geometries, removing the need to create several different geometry files. (Closes #400)

It also implements a method to remove disconnected nodes from the geometry, if it's the case.


vitorslongo:
Volume suppression is already working, however there are still some issues we need to deal with:
~~1. As of now, whenever a volume is supressed, its BCs are simply deleted, with the need to recreate them later if we desuppress it. The most direct solution is to also suppress them, and when we desuppress the volume they reappear as they were created. But this is a whole other issue we already discussed, supressing BCs. Maybe we can deal with this in a separate PR since this is a big change that affects many parts of the code. What do you think, @andrefpf and @JacsonGilVargas? I can't quite determine the size and scope of these modifications.~~
2. I believe with the changes until now we've dealt with half of the problem, because a remesh is needed everytime we suppress some volume (bad, not the way reference softwares do). What I think would be the perfect implementation is: When we already have a generated mesh (every case) and supress something, we change the element matrixes, and not the gmsh mesh, so a remesh is not needed. Then when we actually need to remesh the project for any reason, then we use this `MeshOnlyVisibleEntities` approach, otherwise we would be generating parts of the mesh just to cut off afterwards. The bottom line is: with a generated mesh of everything, we just cut off the mesh of volumes we dont want but already have, and when we do need to generate a new mesh, we do not generate these suppressed parts in the first place. I would also appreciate your opinions on this more complex approach.
~~3. We need a symbol for the volume suppression since it is now on the TreeWidget. (#558)~~

vitorslongo:
> Volume suppression is already working, however there are still some issues we need to deal with:
> 
>     1. As of now, whenever a volume is supressed, its BCs are simply deleted, with the need to recreate them later if we desuppress it. The most direct solution is to also suppress them, and when we desuppress the volume they reappear as they were created. But this is a whole other issue we already discussed, supressing BCs. Maybe we can deal with this in a separate PR since this is a big change that affects many parts of the code. What do you think, @andrefpf and @JacsonGilVargas? I can't quite determine the size and scope of these modifications.
> 
>     2. I believe with the changes until now we've dealt with half of the problem, because a remesh is needed everytime we suppress some volume (bad, not the way reference softwares do). What I think would be the perfect implementation is: When we already have a generated mesh (every case) and supress something, we change the element matrixes, and not the gmsh mesh, so a remesh is not needed. Then when we actually need to remesh the project for any reason, then we use this `MeshOnlyVisibleEntities` approach, otherwise we would be generating parts of the mesh just to cut off afterwards. The bottom line is: with a generated mesh of everything, we just cut off the mesh of volumes we dont want but already have, and when we do need to generate a new mesh, we do not generate these suppressed parts in the first place. I would also appreciate your opinions on this more complex approach.
> 
>     3. We need a symbol for the volume suppression since it is now on the TreeWidget. ([Create symbol for volume suppression #558](https://github.com/MOPT-UFSC/VIBRA/issues/558))

3 was fixed in commit https://github.com/MOPT-UFSC/VIBRA/pull/557/commits/33670430c527d1b933cd7a74c330c51f54b19380. This is the icon I've created (up to suggestions):
image

vibra/engine/mesher/mesh.py
```diff
@@ -76,6 +76,8 @@ def reset_variables(self):
         self.disconnected_nodes_data = dict()
         self.collapsed_elements_data = dict()
 
+        self.suppressed_volumes: set[int] = set()
+
         self.disconnected_nodes = []
```
andrefpf: This list is not necessary anymore. 

We can use directly the `get_disconnected_nodes` method, since it is more precise, simpler and blazingly fast.

vitorslongo:
Now there's a message to the user whenever disconnected nodes exist:
image


To test this, you can comment out this line in `mesh.py` and suppress a volume.
image

vitorslongo:
Making the pending volumes ghost was also implemented:
image

It's ready for review now, @JacsonGilVargas and @andrefpf

vibra/engine/mesher/mesh.py
andrefpf: This comment and related ones are fixed in #562

vibra/engine/mesher/mesh.py
```diff
@@ -1237,13 +1318,27 @@ def post_process_mesh_data(self):
...
+            valid_lines.update(line_tags)
+
         for dim, tag in gmsh.model.getEntities():
             elements_data = dict()
             element_types, element_indexes, element_nodes = gmsh.model.mesh.getElements(dim, tag)
 
             if not element_indexes:
                 continue
 
+            if dim == 2 and valid_surfaces and tag not in valid_surfaces:
```
andrefpf:
```suggestion
            if (dim == 2) and valid_surfaces and (tag not in valid_surfaces):
```
Always use parenthesis to group the logic between ands.
It will save you from weird bugs and it is easier to reason about.

vibra/engine/mesher/mesh.py
andrefpf, line 1: When I open a new project with the same geometry the hidden volumes are kept.
It is probably a bug that also happens in dev, but should be easy to fix. 
A simple "entity_visibility.unhide_all" when creating a new project should do the trick.

(I am commenting on this file just so you can `Resolve conversation`, it is not actually related to this file)

Image

andrefpf, line 1: The new interface is not well fit with the new standard we adopted for all windows.
It should be `cancel apply ok`, with an space between cancel and apply.

Also, the apply callback should not close the window, only ok should do it.

Image

Image

andrefpf, line 1: When I select multiple items and try to remove them, only the last one is removed.

Image

vibra/interface/model_inputs/general/volume_suppression_inputs.py
andrefpf, line 1: There is also this weird bug with selection. Probably some bugged optimization.

https://github.com/user-attachments/assets/6a79f297-fdc9-4cff-9d93-b4c89e8a35fc

andrefpf (changes requested):
The feature is great and well implemented.
But I happen to find a few small easy fix bugs, therefore I will request the changes before merging.

vibra/engine/mesher/mesh.py
vitorslongo: Got it, I've already had to deal with something like this. I'll keep that in mind

vitorslongo:
A new batch of corrections and suggested implementations was made, they include:
- standardize the UI and chage buttons names to "Cancel", "Apply" and "Ok" buttons 
- make "Apply" work accordingly
- change "Delete" and "Add" buttons to "Unsuppress" and "Suppress"
- implement multiple IDs unsupression
- supressing all volumes is now impossible
- now the mesh is not automatically generated, the user is sent to the Mesh setup UI after clicking "Regenerate" for a safer approach
- the suggested element size is now the same as the ones with all volumes unsuppressed, since recalculating yielded errors in some geometries
- selection bug fixed
- supressed volumes visualization filter is now not kept between different loaded geometries

andrefpf (approved):

vibra/interface/data/examples/vibra_files/helicoidal_fluid.vibra
andrefpf, line 1: Please do not modify the example files if not strictly necessary
To revert it I guess you can use the following command:
`git checkout origin/dev -- vibra/interface/data/examples/vibra_files/helicoidal_fluid.vibra`

andrefpf (changes requested):

JacsonGilVargas (approved):
As far as I could verify, everything looks to be working fine. I'll approve this PR as it is, but in the future I'll have some suggestions to make the workflow more user-friendly.

andrefpf (approved):
great work

PR #566, opened 2026-08-31
Add "bad elements" edges to the plot

### LINKED ISSUE #422: Show element edges in bad elements plot

vitorslongo:
This PR fixes the simple problem of the bad elements edges not being plotted. 
This makes it easier to identify what's actually going on th the plot.
Closes #422 

Previously:
image

Now:
image

Other example:
image
image

### DIFF
--- a/vibra/interface/viewer_3d/actors/edges_actor.py
+++ b/vibra/interface/viewer_3d/actors/edges_actor.py
@@ -1,8 +1,9 @@
 from typing import Optional
 
 from molde.colors import Color, color_names
+from vtkmodules.vtkCommonCore import vtkIdList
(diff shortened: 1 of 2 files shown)


andrefpf (approved):
Looks good!

PR #570, opened 2026-09-06
Implement the one-way fluid-structure coupling

### LINKED ISSUE #571: Implement the one-way fluid-structure coupling

In summary, the main objective is to implement the one-way fluid-structure coupling (also known as weak coupling). It would be interesting to restructure the numerical core to also facilitate bidirectional coupling, if possible.

### Mockups / Examples

_No response_

### Contact Details

_No response_

JacsonGilVargas:
This pull request implements all the required code modifications to enable one-way fluid-structure coupling (also known as weak coupling). What is one-way coupling? It's an analysis type in which the acoustic problem is solved first, and the resulting nodal acoustic pressure at the fluid-structure interfaces is then used to compute a structural excitation. Therefore, this is a one-way analysis: acoustic pressure can exert loads on the structural domain, but the acoustic field is not influenced by the structural field. Several code modifications have been implemented since this branch was opened, so only the most relevant will be highlighted here. All element classes (3D, 2D, and 1D) for both acoustic and structural domains were refactored to leverage the class hierarchy, reducing code verbosity. The acoustic and structural assembler classes were also refactored following the same strategy. I'm fully aware that this pull request is insanely big and "more engineering-related", but it's indubitably the most important tool we have already implemented so far. This pain is worth it, since this is exactly what we have been working on since 2023. For the reviewers, I'd appreciate it if you could dedicate some time to precisely reviewing this pull request, since it's the highest priority. Thank you very much in advance.

Closes #571.


andrefpf:
I converted it to draft just to indicate that it should not be tested yet

data/examples/geometry_files/compressor_spool_discharge_first_stage_simp.STEP
andrefpf, line 1: When I executed a coupled analysis for a structure containing only fluid and this error appeared.
I think there are two possible solutions:
- Run normally and display the equivalent of an acoustic analysis
- Raise an error informing that a coupled analysis if a single domain does not make sense

What seems more reasonable @JacsonGilVargas?

Image

andrefpf, line 1: I noticed that it is not possible to add an acoustic property on structural entities.
However, if i set it into a fluid and then replace the fluid with an material, the property is kept.

Image

andrefpf, line 1: Accidentally pressing ok without any selected entities raises an exception in the acoustic pressure dialogs.

Image

andrefpf, line 1: The selected IDs text box is not being updated in the reciprocating compressor dialog 

Image

JacsonGilVargas: Allowing a coupled analysis within a single domain is conceptually incorrect. A robust approach involves extending the analysis checkers to handle coupled analyses and to halt code execution when necessary. I will implement a suitable solution, @andrefpf.

vitorslongo (dismissed):

andrefpf (approved):
As far as I am qualified to test, everything looks great!

vitorslongo (approved):

PR #575, opened 2026-09-15
Updates SectionPlaneWidget's spinboxes behavior

### LINKED ISSUE #488: Section plane does not update plot with spinbox
### What happened?

When a section plane is being configured, changing the values in the spinbox does not update the plot, as seen here:
The Section plane is 90 degrees offset the actual plane on the structure. In other words, only the blue representation of the section plane is being updated.

Image

Also, I believe it would make more sense to change the spinbox step to 5 or 15 degrees, since we do not have the scroolwheel event anymore and going one degree at a time is too little progress.

Windows

### Version

dev branch

### Relevant log output

```shell

```

### Upload screenshots

_No response_

### Contact Details

_No response_

R0drigoSchwartz:
## What does this PR do?

This PR updates the behavior of the `SectionPlaneWidget` spinboxes. Previously, when a spinbox value was changed, the section plane would change position but not apply the cut. Now, the cut is applied whenever the section plane's position is changed using the spinboxes.

This behavior is used only for the spinboxes. The behavior of the sliders has been preserved.

### Demo

https://github.com/user-attachments/assets/ad8f8b84-241a-49b3-8431-fb24833e125d

Closes #488


vitorslongo (approved):
The only thing I noticed is the one from the issue mentioned above: slicing a mesh makes the loading dialog pop up repeatedly. Still, the changes here satisfy this PR's scope, so I'll approve.

andrefpf (approved):

PR #576, opened 2026-09-17
Refactor animation cache

andrefpf:
Refactor the animation caches for efficiency, to fix a bug and to make it clearer.

The dedicated class updates the min and max values according to the data being cached.
Now it is caching the arrays, not the vtk objects, storing less data.
We are also avoiding to store displacement data inside acoustic analysis.


JacsonGilVargas (approved):
Excellent code refactor, @andrefpf.

PR #577, opened 2026-09-18
Refactor show interface behavior

### LINKED ISSUE #504: Loading bar only on slow function

I think we should develop a method to only show the loading bar after like 0.2s.

We usually add the loading bar when the function can be slow, but sometimes, specially in small projects, these functions run pretty fast. 

In these cases it is very annoying to see the a bunch of bars popping and disappearing almost instantly.

### Mockups / Examples

_No response_

### Contact Details

_No response_

GuilhermeSPierri:
This pull request fixes the problem of showing many progress bars frenetically while a function is being executed, regardless of whether it is a fast or a slow function to run.

The solution was to add a `QTimer()` inside the `run()` method, which has a timeout that can be fired to execute the show() method if the function takes more than 200 ms to be executed, otherwise the progress bar interface won't be shown.

Closes #504


vitorslongo (approved):

andrefpf (approved):
It feels much better now

andrefpf:
I increase the delay to 300 ms because in my computer it felt a bit better =)

andrefpf:
I also removed the spinning cursor animation, because Jacson told me to do it a long time ago and I forgot.
The commits put @vitorslongo review on stale, but he had already approved, so feel free to merge.

PR #578, opened 2026-09-18
Billboard symbols

andrefpf:
This PR implements a new type of symbol that combines three variants:
- **Entities** (regular objects that are always in front of everything, such as compressors)
- **Markers** (objects with fixed screen size but relevant direction, such as arrows used in vector boundary conditions)
- **Billboards** (objects with fixed screen size that are always facing the camera, used in scalar boundary conditions, such as impedances, absorption surface and acoustic pressure)

Also removed `symbols_actor_acoustic_fixed_size` from the geometry actor and embedded those symbols in `symbols_actor_acoustic` using the new Entities type.

image


andrefpf:
## Before
image

## After
image

JacsonGilVargas (approved):
I performed a couple of tests and everything is working as expected. Keeping the symbols perpendicular to the user's line of sight helps identify the type of boundary condition without the need to zoom in or extensively manipulate the camera view. Excellent work, @andrefpf.

andrefpf:
Ops, I forgot to wait for the @vini-ribeiro review, sorry. 
If you find something odd you can message me, or create the issue and assign me.

PR #584, opened 2026-09-19
New user preferences window

taianabsfarias:
The new user preferences window has been implemented. The old window showed all settings on a single page, while the new window is split into five categories (Geometry Renderer, Mesh Renderer, Renderer, General, and Subprocess). 
A search bar has also been added that filters items by name and jumps to the category of the respective searched item.
During implementation, i wasnt sure whether to keep the "faces color" button in geometry and mesh, since when checking other files i noticed that the face color is meant to stay fixed (white, or the color of the property assigned to it), so the button ends up having no effect at all, but i decided to keep it anyway. Theres now also a "volumes color" button, which was requested for the mesh tab.
Captura de tela 2026-09-19 142057
Captura de tela 2026-09-19 142445
Captura de tela 2026-09-19 142545
Captura de tela 2026-09-19 142557
Captura de tela 2026-09-19 142605


R0drigoSchwartz (dismissed):
I tested your approach and everything seems to be working fine.

The user preferences looks much cleaner and more organized now. Great job @taianabsfarias 👏

I only suggest you to remove the faces color option, and keep the default color of the volume as white.

vibra/interface/user_preferences.py
```diff
@@ -37,7 +37,7 @@ def set_light_theme(self):
         self.lines_color = color_names.BLACK
         self.edges_color = color_names.BLACK
         self.faces_color = color_names.WHITE
-        self.volumes_color = Color("#4F94CD")
+        self.volumes_color = Color("#FFFFFF")
```
R0drigoSchwartz: For common colors like white, you can use `color_names.WHITE` instead of the `Color("#FFFFFF")`, because it's more readable.

andrefpf (approved):
It looks great

R0drigoSchwartz (approved):
Great job @taianabsfarias 👍🚀

